PR #24, opened 2015-06-10
Updating Scala versions, adding cross-compilation

malcolmgreaves:
-- scala version is now 2.11.6
-- Added cross-compilation scala versions (2.11.6, 2.10.5)
-- Updated version number to 5.3.1

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,10 +1,12 @@
 name := "processors"
 
-version := "5.3-SNAPSHOT"
+version := "5.3.1-SNAPSHOT"
 
 organization := "edu.arizona.sista"
 
-scalaVersion := "2.11.5"
+scalaVersion := "2.11.6"
+
+crossScalaVersions := Seq("2.11.6", "2.10.5")
 
 scalacOptions ++= Seq("-feature", "-unchecked", "-deprecation")
 


malcolmgreaves:
With this, one can publish using `sbt +publish` to get a 2.10 and a 2.11 version.

MihaiSurdeanu:
Hi Malcom,
We are in middle of a big deadline. We will pull in a week or so.
Btw, scala 2.10 might not work anymore due to a new dependency. We'll check.
Thank you!
Mihai

marcovzla:
Hi Malcolm,

Thanks for the pull request. Cross compiling for scala 2.10 would be very useful indeed.
We are just wondering if this solution is enough, because we suspect there are two issues:

1. We believe we must include some dependencies conditionally based on the scala version. Particularly, parser-combinators where part of the standard library in 2.10 but need to be included in 2.11.
2. There is a component of the Odin framework that relies heavily on reflection. As far as I know, reflection wasn't thread safe in 2.10, but it is in 2.11. This is a big deal, because we need to use odin in parallel. A solution would be to change the architecture so that we don't use reflection, but it may take some time.

Let us know if you have any ideas about these issues. We will try to address these issues but it may take some more time.

malcolmgreaves:
Hi Marcovzla,

Thanks for a thorough explanation of the (probable) problems with compiling for 2.10! I'm using this library plus Spark in a project I have. It's rather unfortunate that, even as of 1.4, Spark doesn't cross compile to 2.11.

In regards to the problems:

(1) I think we could follow the approach of this SO Q&A: http://stackoverflow.com/questions/12626197/conditional-scalacoptions-with-sbt We could match on the scalaVersion during build and modify the dependencies to either include or exclude parser-combinators.

(2) Ah, that seems like quite a hairy beast... In a different project, I ran into a similar problem with runtime inspection of generic types using TypeTags in 2.10. Had to refactor to use Manifests since I had to use Spark.

I wonder if it's worth the effort to try and support 2.10? Maybe just bug the Spark people to release a 2.11 =P

marcovzla:
I will close this for now. Odin requires Scala 2.11, but we may take it out of processors, which would make cross-compiling to 2.10 easier.

malcolmgreaves:
Sounds good. Thanks for reviewing!

PR #38, opened 2016-02-18
Added chunks to printDoc

### LINKED ISSUE #36: Add chunks to ProcessorExample
[ProcessorExample](https://github.com/clulab/processors/blob/299cccba73df83ef7033fee43da2a185a21af395/src/main/scala/edu/arizona/sista/processors/ProcessorExample.scala#L44-L48) should be updated to display chunks, if present.

myedibleenso:
I've added the display of chunks to the `printDoc` method of the `ProcessorExample`, which is also used by the `processorShell`.

The change resolves #36.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,13 +55,14 @@ This software is available on Maven Central. To use, simply add the following de
        <classifier>models</classifier>
     </dependency>
 
- The equivalent SBT dependencies are:
+The equivalent SBT dependencies are:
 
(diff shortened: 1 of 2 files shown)


myedibleenso:
I've also updated the README to reflect the change to the example.  

This last change is purely cosmetic, but the fenced code blocks in the README now use language-specific syntax highlighting.

PR #39, opened 2016-02-19
use matched tokenInterval when available

marcovzla:
This pull request changes the behavior of `TokenExtractor` so that the matched `tokenInterval`
is used even when the result is an `EventMention` or `RelationMention`.

This is more intuitive behavior than the current, in which only the span of the captured arguments is considered.


marcovzla:
this still needs some tests

myedibleenso:
I'll try to add some tests later today.

myedibleenso:
Tests look good!  I'm going to merge this.

PR #40, opened 2016-02-20
use DependencyUtils.findHeadsStrict

marcovzla:
This pull request fixes a bug in Mention.synHead* methods so that a preposition is not considered a head

### DIFF
--- a/src/main/scala/edu/arizona/sista/odin/Mention.scala
+++ b/src/main/scala/edu/arizona/sista/odin/Mention.scala
@@ -99,10 +99,7 @@ trait Mention extends Equals with Ordered[Mention] with Serializable {
       // we don't need dependencies, a single token is its own head
       tokenInterval
     } else {
-      sentenceObj.dependencies match {
-        case Some(deps) => DependencyUtils.findHeads(tokenInterval, deps)
-        case None => Nil
-      }
+      DependencyUtils.findHeadsStrict(tokenInterval, sentenceObj)
     }
   }
 


marcovzla:
tests needed

MihaiSurdeanu:
But a IN is actually the head of a PP...
See Michael Collins' PhD thesis appendix.

myedibleenso:
Another reason why we don't always want the syntactic head (for example, the causal relations project makes some choices based on properties of the selected head.  We probably want to be looking at something other than the head here).  In any case, `findHeadsStrict` is better than what we're using currently in that it ignores punctuation.  

I think a better solution might be to remove the "cannot be a prep" constraint from this method, or rather make it a soft constraint on the head selection policy (i.e., the function for picking one head whenever more than one is found for a span).  We could try something like picking the first/last head that is not a prep whenever at least one prep-less head is found.  If only preps are found, we could fall back to the current policy (choose the last?).

MihaiSurdeanu:
I would prefer the Stanford strategy, which implements multiple models for finding the head of phrase. They have the "classic" Collins algorithm. But also something called "findSemanticHead", which does exactly what we discuss here, i.e., skipping INs when looking for the head of a PP (and a few other small things). 
I can see how both would be needed depending on the task. I think we should go for that approach.

marcovzla:
`findHeadsStrict` removes punctuation, but also removes some words based on POS tag, which sounds like it is not what we want: `val stopTags = "(.|,|\\(|\\)|:|''|``|#|$|CC|TO|IN)"`

MihaiSurdeanu:
I think this needs cleaning up. The code should be split in two distinct methods:
- findSyntacticHead - follows the standard algorithm for syn deps. This is useful for people working on syntax.
- findSemanticHead - this could be the same algorithm as findHeadsStrict. This is a more semantic-y interpretation of head words, which is useful for people working on IE.

marcovzla:
I'm closing this pull request. A better solution is in pull request #45

PR #42, opened 2016-02-28
Improved random forest classifier; added feature selection by IG; Datasets are now serializable.

MihaiSurdeanu:


src/main/scala/edu/arizona/sista/learning/Dataset.scala
```diff
@@ -89,26 +92,86 @@ class BVFDataset[L, F] (
...
     val counts = new Counter[Int]
     for(d <- fs) {
       for(f <- d) {
         counts.incrementCount(f)
       }
     }
-    counts
+    logger.debug("Total unique features before filtering: " + counts.size)
+
+    val passed = new mutable.HashSet[Int]()
```
myedibleenso: If you don't want to have to handle those conversions (i.e., mutable to immutable), you could go with something like this:

```scala
val passed: Set[Int] = for {
  f <- counts.keySet
  if(counts.getCount(f) >= threshold)
} yield f
```

src/main/scala/edu/arizona/sista/learning/Dataset.scala
```diff
@@ -338,6 +344,32 @@ class RVFDataset[L, F] (
...
+      }
+      neg *= pWithout(total)
+    }
+    pos + neg
+  }
+
+  def pWith(total:InformationGain) = datumCount.toDouble / total.datumCount.toDouble
+  def pWithout(total:InformationGain) = (total.datumCount - datumCount).toDouble / total.datumCount.toDouble
+}
+
```
myedibleenso: Sorry, I'm not that familiar with this project so I'm probably missing something obvious.  Why does `datumCount` need to be a `var`?  It doesn't look like its value is reassigned anywhere in here.

src/main/scala/edu/arizona/sista/learning/RFClassifier.scala
```diff
@@ -628,7 +667,10 @@ case class Utility (feature:Int, // index of this feature in the feature lexicon
...
 class RFJob[L, F](
                    val dataset:CounterDataset[L, F],
-                   val indices:Array[Int],
+                   val trainIndices:Array[Int],
+                   val oobIndices:Array[Int],
+                   val activeNodes:Set[(Int, Double)],
+                   val nilLabel:Option[L],
                    val featureThresholds:Array[Array[Double]],
                    val currentUtility:Double,
                    val random:Random) {
```
myedibleenso: @marcovzla has been trying to get us all to follow these conventions: http://docs.scala-lang.org/style/indentation

What do you think?

src/main/scala/edu/arizona/sista/learning/Dataset.scala
MihaiSurdeanu: Yes, this is more elegant than my code. I didn't know you could assign a type to the seq produced by yield.

src/main/scala/edu/arizona/sista/learning/Dataset.scala
MihaiSurdeanu: It is incremented somewhere else. This is one of the few mutable objects that should be mutable...

src/main/scala/edu/arizona/sista/learning/RFClassifier.scala
MihaiSurdeanu: I agree with the 2-space indentation! This must've slipped through the code...
Good comments!

PR #44, opened 2016-03-05
findHeads() handles graph cycles

### LINKED ISSUE #43: Recursion error in DependecyUtils
I found an error when using the DependencyUtils.  It seems like it's getting stuck in a loop in the dependency graph and the recursion blows the stack.  This is a snippet of the stack trace:

```
Caused by: java.lang.StackOverflowError
    at scala.runtime.BoxesRunTime.boxToInteger(BoxesRunTime.java:69)
    at edu.arizona.sista.processors.Sentence.dependencies(Document.scala:73)
    at edu.arizona.sista.utils.DependencyUtils$.getIncoming$2(DependencyUtils.scala:129)
    at edu.arizona.sista.utils.DependencyUtils$.followTrailOut$2(DependencyUtils.scala:145)
    at edu.arizona.sista.utils.DependencyUtils$.edu$arizona$sista$utils$DependencyUtils$$followTrail$3(DependencyUtils.scala:138)
    at edu.arizona.sista.utils.DependencyUtils$.followTrailOut$2(DependencyUtils.scala:150)
```

marcovzla:
This pull request reimplements `DependencyUtils.findHeads()` to handle cycles in the dependency graph.

It also reimplements `DependencyUtils.findHeadsStrict()` to call `findHeads()` instead of reimplementing graph traversal.

Fixes #43

### DIFF
--- a/src/main/scala/edu/arizona/sista/utils/DependencyUtils.scala
+++ b/src/main/scala/edu/arizona/sista/utils/DependencyUtils.scala
@@ -85,21 +85,19 @@ object DependencyUtils {
       case (tok, dist) :: rest if seen contains tok =>
         // we already explored this token, skip
         countSteps(rest, seen)
+      case (tok, dist) :: rest if graph.roots contains tok =>
+        // found a root
+        // it is the closest one because we are searching breath-first
+        // return distance
+        dist
       case (tok, dist) :: rest =>
         // explore
         val incoming = followIncoming(tok, graph)
         if (incoming.isEmpty) {
-          if (graph.roots contains tok) {
-            // found a root
-            // it is the closest one because we are searching breath-first
-            // return distance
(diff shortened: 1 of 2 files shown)


marcovzla:
Seems like this solution is still incomplete. Please don't merge yet.

marcovzla:
There was still an issue if the root node had incoming edges, which sounds wrong to me but apparently it is possible. Added a test and a fix for that situation.

myedibleenso:
Yeah, that sounds wrong to me, too.  Why would the root have anything incoming? It sounds like a red flag that the parse is wrong. Anyway, at least you're handling those cases.

MihaiSurdeanu:
It can happen because the parser is dumb. It makes local decisions wo/ worrying about global coherence.

PR #45, opened 2016-03-05
added semHead* methods to Mention

marcovzla:
This pull requests adds `semHead` methods to `odin.Mention` to retrieve a mention's semantic head. They are implemented as a call to `DependencyUtils.findHeadsStrict()`.

`synHead` methods are still available. They retrieve the syntactic head. They are implemented as a call to `DependencyUtils.findHeads()`.


myedibleenso:
I think that `DependencyUtils.findHeadsStrict()` should perhaps be renamed to DependencyUtils.findSemanticHeads()`.  Opinions?

MihaiSurdeanu:
I could go either way. findHeadsStrict is a more honest name, that doesn't make any claim about semantics :)

PR #46, opened 2016-03-05
throw exception when dependency parse when root is unreachable

marcovzla:
We found a sentence in the agiga corpus that has a messed up dependency parse.
The issue is that there is a loop (a node pointing to itself) that makes the root unreachable.

This pull request introduces a `DependencyUtilsException` that is thrown when that happens.


MihaiSurdeanu:
I'll merge this because this is a valid test. But we (or I...) should modify the parser in processors to never allow this to happen.

marcovzla:
this is not our parser. It is a sentence from the agiga corpus. I sent you an email with more details.
I can paste it here if you want to

MihaiSurdeanu:
Agreed. But they used CoreNLP to parse this, afaik. So it could happen again.

PR #64, opened 2016-06-15
Sentiment analysis support in CoreNLPProcessor

myedibleenso:
These changes add (and test) support for tree-based sentiment analysis in `CoreNLPProcessor`.


MihaiSurdeanu:
I would prefer not to merge this in master. See my email for a simpler solution.

PR #65, opened 2016-06-15
Sentiment Analysis support

myedibleenso:
These changes add tested support for sentiment analysis via a trait (`SentimentAnalyzer`) and an interface to CoreNLP's tree-based sentiment analysis in `CoreNLPSentimentAnalyzer`.


MihaiSurdeanu:
The SentimentAnalyzer is never used. Is it really needed?

myedibleenso:
`CoreNLPSentimentAnalyzer` makes use of this trait.  It isn't essential, but I included the trait to enforce a uniform API for sentiment analysis in the future.  I imagine others may add different implementations for sentiment analysis.  It would be nice if these are called the same.

MihaiSurdeanu:
merged. I will release 5.8.6 tonight, which will include this.

PR #76, opened 2016-07-11
Ner override

hickst:
Trying to get this branch into our next DryRun (today).


myedibleenso:
This looks like some convenience methods + support for NER override.  Is that an accurate summary?

PR #80, opened 2016-07-21
Distributional similarity comparisons for Odin 

myedibleenso:
These changes add the following functionality to Odin:

1. Distributional similarity comparisons using word embeddings  
    a) includes a resource manager for `Odin`.  Right now the manager is only being used to load a set of embeddings via the [`Word2vec` class](https://github.com/clulab/processors/blob/9ce0bcc59bfbcbc0ef9a982644224d944391effa/src/main/scala/org/clulab/embeddings/word2vec/Word2vec.scala), but it designed to be extensible so that we could handle a wordnet db, etc.   
    b) `simScore(word)` calculates the cosine similarity of the embedding for some word (`word`) and the current token being matched.  It returns a floating point value.
2. Numeric comparisons:
```yaml
[2+2 < 5] # will always match a token (at least in this universe)
```

An example of the new feature:

```yaml
# find a token that is more similar to "cat" than it is to "dog", 
# where its cosine similarity with "cat" is >= 0.4
[(simScore("cat") >= 0.4) & (simScore("cat") >= simScore("dog"))]
````

The embeddings file to be used must be specified in the rule file in the following manner:
```
resources:
  # a file containing embeddings with a one-line header that provides the dimensions of the embeddings and the size of the vocabulary
 # this file format is identical to what is produced by training a model with word2vec
  embeddings: org/clulab/odin/embeddings/w2v-gigaword-sample.txt
```

In the future I would like to add expressions to compose embeddings for comparisons:
```yaml
# the current token must be similar to vec(king) - vec(man) + vec(woman)
[simScore(embedding("king") - embedding("man") + embedding("woman")) >= 0.5]
```
In order to do that, though, we need to decide on an implementation for a vector.


MihaiSurdeanu:
This is awesome.

PR #86, opened 2016-07-26
WIP better support for multitoken triggers

marcovzla:
This pull request changes the way multi-token triggers are handled by odin.
Previously, odin would try each token in the trigger as a possible starting point for the dependency patterns, but odin required all the arguments to have the same starting point.
Now different syntactic paths can have different starting points inside the trigger, allowing us to write more flexible rules.

### DIFF
--- a/src/main/scala/org/clulab/odin/impl/DependencyPattern.scala
+++ b/src/main/scala/org/clulab/odin/impl/DependencyPattern.scala
@@ -22,44 +22,69 @@ trait DependencyPattern {
       ruleName: String
   ): Seq[Mention]
 
-  // extract the arguments of a token interval
   protected def extractArguments(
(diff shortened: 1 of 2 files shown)


marcovzla:
@MihaiSurdeanu this should fix the failing tests in the feedback2 branch of reach

myedibleenso:
Great.  We should try to come up with a test or two for this that would be appropriate for processors.

myedibleenso:
Unfortunately, this improvement is breaking a few tests in reach.  We'll look into it.  In any case, we should probably wait to merge until we have tests for this.

myedibleenso:
1. All tests pass on the `master` branch of `reach` and `processors`
2. @marcovzla added a test

I think this should be merged.

PR #90, opened 2016-09-06
json serialization

### LINKED ISSUE #89: Serialize/deserialize Mentions to/from json
We really need to add support to write/read mentions to/from json.

myedibleenso:
This pull request resolves #89.

## Summary of changes

- added `org.clulab.serialization.json` package containing implicit class conversions with `json` serialization\deserialization support for `Document`, `Sentence`, and Mentions.
  - Design avoids redundancies of including a nested `json` representation of the `Document` associated with each `Mention`.  Instead, the `Mention`'s `json` contains a pointer to its `Document` (see attached example). 
- simplified constructors for `Document` and `Sentence`
- added support for comparing contents of `Sentence`, `Document`, and `DirectedGraph` that can be used for establishing equivalence of representations for annotated text.
- some refactoring

Complete working example demonstrating `json` methods:

```scala
import org.clulab.processors.fastnlp.FastNLPProcessor
import org.clulab.processors.Document
import org.clulab.struct.Interval
import org.clulab.odin._
import org.clulab.serialization.json._
import java.io.File

val rules =
  """
    |# NE rules
    |
    |- name: "ner-person"
    |  label: [Person, PossiblePerson, Entity]
    |  priority: 1
    |  type: token
    |  pattern: |
    |   [entity="PERSON"]+
    |
    |# Events
    |
    |# optional location and date
    |- name: "marry-syntax-1"
    |  label: [Marry]
    |  priority: 3
    |  example: "He married Jane last June in Hawaii."
    |  type: dependency
    |  pattern: |
    |    trigger = [lemma="marry"]
    |    spouse: Entity+ = <xcomp? /^nsubj/ | dobj
  """.stripMargin
val engine = ExtractorEngine(rules)

val proc = new FastNLPProcessor
val text = "Gonzo married Camilla."
val doc = proc.annotate(text)
val mentions = engine.extractFrom(doc)

// print the json string for the extracted mentions
println(mentions.json(pretty=true))

// save the json to a file
mentions.saveJSON("mentions.json", pretty=true)

// deserialize the mentions json that was written in the last step
val mentions2 = JSONSerializer.toMentions(new File("mentions.json"))

// demonstrate span equivalence to the original mentions
def m2triple(m: Mention):(Int, Int, Interval) = (m.document.equivalenceHash, m.sentence, m.tokenInterval)
mentions.map(m2triple) == mentions2.map(m2triple)

// documents can also be serialized/deserialized
doc.saveJSON("document.json", pretty=true)
val d2 = JSONSerializer.toDocument(new File("document.json"))

// demonstrate equivalence of annotations
doc.equivalenceHash == d2.equivalenceHash
```

The "document.json" produced in the example above is attached here
[document.json.txt](https://github.com/clulab/processors/files/456053/document.json.txt)
.

### DIFF
--- a/src/main/scala/org/clulab/discourse/rstparser/RelationFeatureExtractor.scala
+++ b/src/main/scala/org/clulab/discourse/rstparser/RelationFeatureExtractor.scala
@@ -1,8 +1,7 @@
 package org.clulab.discourse.rstparser
 
-import org.clulab.processors.{CorefMention, Document}
-import org.clulab.struct.Counter
-import org.clulab.struct.Tree
+import org.clulab.processors.Document
+import org.clulab.struct.{CorefMention, Counter, Tree}
 import Utils._
 
 /**
--- a/src/main/scala/org/clulab/odin/Mention.scala
+++ b/src/main/scala/org/clulab/odin/Mention.scala
@@ -6,9 +6,7 @@ import org.clulab.struct.Interval
 import org.clulab.processors.Document
 import org.clulab.utils.DependencyUtils
 import org.clulab.odin.impl.StringMatcher
-import org.json4s._
-import org.json4s.JsonDSL._
-import org.json4s.native._
+
 
 @SerialVersionUID(1L)
 trait Mention extends Equals with Ordered[Mention] with Serializable {
@@ -135,12 +133,6 @@ trait Mention extends Equals with Ordered[Mention] with Serializable {
       bits.mkString
   }
 
-  def jsonAST: JValue
-
-  def json(pretty: Boolean = false): String =
-    if (pretty) prettyJson(renderJValue(jsonAST))
-    else compactJson(renderJValue(jsonAST))
-
   override def canEqual(a: Any) = a.isInstanceOf[Mention]
 
   override def equals(that: Any): Boolean = that match {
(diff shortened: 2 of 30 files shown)


hickst:
I like it in general. But it seems like "stanford-basic" and "stanford-collapsed" let the implementation bleed through, which may change in future. How about "basic" and "collapsed", (which also shorten the output size)?

hickst:
Is there an option to produce the original text as part of the Document JSON output? I don't see the text in the given small example. And the Document ID will also be produced, if there is one?

MihaiSurdeanu:
@myedibleenso: can you please add a summary of this PR to CHANGES?

myedibleenso:
>Is there an option to produce the original text as part of the Document JSON output? 

The reason why you're not seeing it in the attached example is that the `.annotate` call in the example above does not use `keepText = true` (default is false).  If the text is set for the `Document`, it will appear in the `json` and be added to the new `Document` during deserialization (would be good to add a test or two for this though).

>And the Document ID will also be produced, if there is one?  

The same thing goes for the `id` (it will be serialized whenever present and then deserialized)

>I like it in general. But it seems like "stanford-basic" and "stanford-collapsed" let the implementation bleed through, which may change in future. How about "basic" and "collapsed", (which also shorten the output size)?  

While I like shrinking the output size, this (hopefully) takes out the guesswork of what representation the graph uses.  In any case, it is whatever keys are used for the `GraphMap`.  Actually, the contents of the `GraphMap` can be changed/updated even after annotation. We could change the default strings used by that Map, though (I think that "collapsed" and "basic" are too terse.  I could imagine a different representation with a "collapsed" variant).

>@myedibleenso: can you please add a summary of this PR to CHANGES?

Sure.  Sorry, I forgot to include that in the PR when I opened it.  Should I just push that directly to `master` or open another PR?

PR #98, opened 2016-10-18
Update README.md

moldovean:
1. the method was waiting for an int but a boolean was provided instead.
2. the import org.clulab.struct.CorefMention was missing 
regards 4m Moldova


MihaiSurdeanu:
Thanks Adrian!

Can you please change that 0 to ShallowNLPProcessor.NO_DISCOURSE, for consistency with other code?

PR #99, opened 2016-10-20
Odin support for Event.arg and Event.trigger patterns

myedibleenso:
## `Event.argument` and `Event.trigger` patterns

These changes add support for patterns over a mention's arguments and trigger.

For example, consider the following sentence:
>MEK bound Ras binds with Kras

For the sake of the example, suspend your disbelief of the biology and let's imagine that a Binding(MEK, Ras) and Binding(Ras, Kras) have been observed.

We can now write patterns to inspect the arguments and/or triggers of mentions:

`@Binding.theme @trig1:Binding.trigger @Binding.theme @trig2:Binding.trigger`
`@Binding.theme [mention=Binding.trigger & tag=VBD] @Binding.theme [mention=Binding.trigger & tag=VBZ]`

A match by the first pattern will produce a `RelationMention` with the named arguments `trig1` and `trig2`, corresponding to the binding triggers "bound" and "bind" in the example sentence.

The second pattern demonstrates how these pattern extensions can be used in token constraints through the `mention` field in conjunction with constraints over other token attribute fields.

Tests are here: https://github.com/clulab/processors/blob/mentionArgs/main/src/test/scala/org/clulab/odin/TestTokenPattern.scala#L793-L898


myedibleenso:
We've also moved the odin tests to the `main` subproject and now deserialize the annotated documents used in those tests to avoid a dependency on the `corenlp` subproject for the tests in `main`.

myedibleenso:
It's also worth mentioning that we don't support something like `Regulation.controlled.controlled.theme`.

MihaiSurdeanu:
@myedibleenso: please update CHANGES.md for this PR.

PR #100, opened 2016-10-23
recognize overlapping entities with differing starts

### LINKED ISSUE #82: Overlapping entities with differing starts are not recognized
With a rule such as [this](https://github.com/clulab/reach/blob/a32a3ef7eb94322cb13fea0a5bba452959b83347/src/main/resources/org/clulab/reach/biogrammar/events/pos-reg_template.yml#L283-L290), a sentence such as `ASPP1 and ASPP2 are phosphorylated in response to EGFR` results in only one regulation. Both phosphorylation events `[ASPP1 and [ASPP2 are phosphorylated]]` should be found as controllers by this rule. Only the longer event is found.

marcovzla:
Overlapping entities with different starts were not handled correctly when at the beginning of a pattern. This PR implements cautious advance when the first match in a surface pattern is a MatchMention, but not for any other kind of match.

Closes #82

### DIFF
--- a/main/src/main/scala/org/clulab/odin/impl/TokenPattern.scala
+++ b/main/src/main/scala/org/clulab/odin/impl/TokenPattern.scala
@@ -20,9 +20,9 @@ object TokenPattern {
   def startsWithMatchMention(inst: Inst): Boolean = inst match {
     case i: MatchMention => true
     case i: Pass => startsWithMatchMention(i.next)
-    case i: Split => startsWithMatchMention(i.lhs) || startsWithMatchMention(i.rhs)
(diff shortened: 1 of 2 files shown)


marcovzla:
This PR breaks this Reach test: https://github.com/clulab/reach/blob/c8f733d1cba5b13b0d9f87362d3f6229262ebfb7/main/src/test/scala/org/clulab/reach/TestTokenPatterns.scala#L35-L43

I believe that test is wrong and need to be modified. Particularly, those textboundmentions should be eventmentions with the same span but different parameters.

MihaiSurdeanu:
@marcovzla: I agree that test is incorrect. Please adjust.

MihaiSurdeanu:
@marcovzla: please update CHANGES.md for this PR.

Issue #68, opened 2016-06-16
Redesign odin.Mention to properly address cross-sentence cases


myedibleenso:
`odin.Mention` does a poor job of handling cross-sentence cases.  While it requires some careful thought, we should plan on rewriting this class to better handle cross-sentence mentions (important for coref among other things).

PR #111, opened 2016-11-03
Cross sentence matcher

myedibleenso:
These changes introduce a cross-sentence odin extractor (denoted via `type: "cross-sentence"`) and a `CrossSentenceMention` that links two existing mentions within a sentential window (specified using `right-window: someInt` and `left-window:someInt`).  This allows for patterns that "filter" existing mentions matching the specified type.  I believe that this matcher may be useful for coreference resolution and entity tracking throughout a document.

Note that a pair of mentions occurring in the same sentence will not be linked by a `cross-sentence` pattern.  The new `CrossSentenceMention` partially addresses #68.  In the future, we might consider a requirement in `RelationMention` and `EventMention` to ensure that the arguments (and trigger) are in the same sentence, but this might call for an adjustment to the state.  

tests are here:
https://github.com/clulab/processors/blob/e5f3d2df1a53b3363e603aa35a76ffdf5f8c5f6e/main/src/test/scala/org/clulab/odin/TestCrossSentencePatterns.scala

examples in reach:
https://github.com/clulab/reach/blob/4a5e876ec8cffb8790c188abe083d7ea04e0ce33/assembly/src/main/resources/org/clulab/reach/assembly/grammars/intersentential.yml


MihaiSurdeanu:
@myedibleenso: Please update CHANGES.md.

MihaiSurdeanu:
@myedibleenso, @marcovzla, @hickst : Ok to release processors?

MihaiSurdeanu:
@myedibleenso: ideally, we shouldn't need a separate CrossSentenceMention. The default Mention class should transparently handle within- and cross-sentence spans... We can discuss this at some point.

PR #121, opened 2016-12-09
BioNLPProcessor now inherits from FastNLPProcessor

myedibleenso:
`BioNLPProcessor` now extends `FastNLPProcessor instead of `CoreNLProcessor`.  `FastNLPProcessor` uses a parsing model that is both faster and performs ~12 pts (F1) better than `CoreNLPProcessor` the on GENIA test data (performance on the GENIA corpus was measured and reported by @terroni).

### DIFF
--- a/corenlp/src/main/scala/org/clulab/processors/bionlp/BioNLPProcessor.scala
+++ b/corenlp/src/main/scala/org/clulab/processors/bionlp/BioNLPProcessor.scala
@@ -4,10 +4,10 @@ import java.util
 import java.util.Properties
 import java.util.regex.Pattern
 
-import org.clulab.processors.bionlp.ner.{KBLoader, BioNER, RuleNER}
+import org.clulab.processors.bionlp.ner.{BioNER, KBLoader, RuleNER}
 import org.clulab.processors.shallownlp.ShallowNLPProcessor
-import org.clulab.processors.{Sentence, Document}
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
have you tried running the reach tests with this?

myedibleenso:
No.  There are two failing tests anyway, so I guess I will have to do some spelunking.

myedibleenso:
@MihaiSurdeanu, I've fixed the failing tests in `processors`.  The bad news is that when I ran the `reach` tests with this change, over 30 tests fail.  

One thing I've noticed after comparing parses is that with the switch to the `FastNLPProcessor` parser in `BioNLPProcessor`, we're now finding (many) [prepositional clausal modifiers (`prepc_`)](http://nlp.stanford.edu/software/dependencies_manual.pdf).  

There are **31** tests in `reach/main` (run via `sbt main/test`), 2 in `reach/assembly` (run via `sbt causalAssembly/test`), and 1 in `reach/export` (run via `sbt export/test`).   In some cases we're finding additional mentions, while in others we're finding fewer than expected (usually none).

It will require some effort to fix these issues, but it is hard to estimate how much.   For example, we will need to adjust the rules to allow for clausal modifiers (ex. selectively replacing `/prep_/` with `/prepc?_/`).

MihaiSurdeanu:
Thanks!

Hmm… This makes me wonder if we should just bite the bullet and switch to universal dependencies...

Issue #124, opened 2017-01-01
Enhance NER to load multiple override files


hickst:
I've generated a KB file (Phase3-Override.tsv.gz) for the proteins specified by MITRE for Phase3. It contains almost 700 entries, some of which probably conflict with our existing KBs and NER override file.

While we could add all these entries to the front of the existing NER override file (more than doubling its size), it would be better if Reach could handle multiple override files. The Grounding package can already handle this (reading the list of override files from the application.conf file) but the NER needs to be enhanced to load and process multiple override KBs (preferably from a configuration list).

This would also be extremely valuable for "one-off" processing requests from various groups who wish to use their own, custom groundings which override our default mappings.

MihaiSurdeanu:
@hickst: can you please implement this?

We need to modify the `loadAll` method in KBLoader:
https://github.com/clulab/processors/blob/master/corenlp/src/main/scala/org/clulab/processors/bionlp/ner/KBLoader.scala

hickst:
This is implemented and passes Processor tests on my machine but runs out of memory on Travis.

MihaiSurdeanu:
Ok. Then I propose to merge and release processors.
We will have to run some experiments for CMU in the next couple of days for the next DARPA eval. So we should have this available.

hickst:
I'm retesting with latest right now, just to be sure.

hickst:
Tests pass again BUT let's hold off on merge and release until we discuss -- there are other issues which may need to go into a new release.

hickst:
Code was merged by pull request #129

PR #128, opened 2017-01-11
Issue #124: enhance NER to load Phase3 override KB

hickst:
Enhance NER to load multiple override files. Update to use Bioresources 1.1.20 containing MITRE PhaseIII override KB.


MihaiSurdeanu:
This is good, but I wonder if we can come up with a naming convention for the override files. Phase3-Override.tsv.gz seems very arbitrary... Maybe we can have NER-Grounding-Override-ddd.tsv.gz, where the ddd indicates the priority in loading (the smaller the number the higher the priority)?

hickst:
Sure, that sounds fine but there are some other considerations with this file that we should discuss at our weekly meeting tomorrow.

hickst:
I just re-merged master into this branch **after** starting the pull request, then pushed again. I am hoping that Travis will do the right thing and restart the PR tests?

hickst:
Restart Travis testing.

PR #129, opened 2017-01-11
Enhance NER to load Phase3 override KB - 2nd try

hickst:
Enhance NER to load multiple override files. Update to use Bioresources 1.1.20 containing MITRE PhaseIII override KB.

Second try to get Travis to run tests.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,3 @@
--J-Xmx7G
+-J-Xmx6G
 -J-XX:+UseG1GC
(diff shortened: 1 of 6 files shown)


hickst:
Tests are continuing to fail with insufficient memory:

```
There is insufficient memory for the Java Runtime Environment to continue.
Native memory allocation (mmap) failed to map 2042626048 bytes for committing reserved memory.
```

MihaiSurdeanu:
Let's discuss today.

PR #132, opened 2017-01-12
Update to use Bioresources 1.1.21 

hickst:
…containing updated MITRE Phase3 override KB (sites truncated from entries).


hickst:
Reopen to restart Travis.

PR #135, opened 2017-02-08
Remove Phase3 override file to phase3 branch only.

hickst:
We agreed not to depend on MITRE Phase3 override files since they conflict with other collaborator's judgements, but the master branch was accidentally contaminated some time back. This change relegates the Phase3 override file to the `phase3` branch only.


hickst:
Travis running out of memory during the tests again. All tests passed for me running with 8G.

PR #138, opened 2017-02-14
Update to use newly released Bioresources 1.1.22.

hickst:


MihaiSurdeanu:
build fails again...

hickst:
Yes. It's the usual memory problem. What can we do to prevent this from happening every time? I almost always run the exact same test on my machine in 8G before I submit a PR (I did this time, too -- and it passed with no problems).

MihaiSurdeanu:
I wonder if we can implement that simple trick we discussed: filter out proteins (and chemicals?) that are not mentioned at all in Open Access.

hickst:
Do you have some information that points to the KBs as the problem? I have no such information and the Reach tests all pass without running out of memory so I would say it is more likely that the Processor tests themselves are using too much memory.

MihaiSurdeanu:
processors tests are smaller than reach’s.
But you are welcome to take them out (one at a time) to see where it fails.

PR #146, opened 2017-04-22
Iss499: author name confused with protein family

hickst:
Add logic to remove bibliographic references in parentheses.


MihaiSurdeanu:
@hickst: this is way to aggressive: you are removing any statements within ().
We need more informed patterns, to guess if what's inside looks like a citation. For example, does it contain "et al."? Does it contain a 4-digit year? Does it have that XREF_BIB tag?

hickst:
That is incorrect. Please read the code again.

MihaiSurdeanu:
I see. I missed the stringHasBibRef() call. But why isn't that functionality inside the pattern anyway? Seems overkill the way it looks now.

hickst:
It could be but the test is a logically separate function; especially in case we want to complicate the test in the future.

PR #147, opened 2017-04-22
Add tests for bibliographic reference removal (issue 499).

hickst:
Added some tests to validate issue 499 modifications.


MihaiSurdeanu:
Great!

PR #151, opened 2017-06-01
Odin subproject

marcovzla:
Made a subproject for odin to manage its dependencies separately.


MihaiSurdeanu:
Can you add this info to CHANGES?

PR #154, opened 2017-06-22
Ranged quantifiers for graph (dependency) pattern arguments

### LINKED ISSUE #148: Range quantifier error
```
    - name: "marriage-event"
      label: Marriage
      pattern: |
          trigger = [lemma=marry]
          spouse: Person{2} = dobj
```
works but
```
    - name: "marriage-event"
      label: Marriage
      pattern: |
          trigger = [lemma=marry]
          spouse: Person{2,4} = dobj
```
gives me the following error:
```
OdinError: Error parsing rule 'marriage-event': `}' expected but `,' found
```

myedibleenso:
These changes resolve #148 and add support for `{s,e}`, `{,e}`, and `{s,}` ranged quantifiers to arguments in graph (dependency) patterns.  Tests are included.

### `{s,e}`
- The argument must be matched a minimum of `s` times and a maximum of `e` times.  If more than `e` Mentions are found to match the argument pattern, a mention for each combination of matches of size `e` will be produced.  
   - `argName: ArgType{1,3} =` with **2** matches will produces **1** (event/relation) mention with **2** mentions for the argument key,`argName`.
   - `argName: ArgType{1,3} =` with **1** match will produces **1** (event/relation) mention with **1** mention for the argument key,`argName`.
   - `argName: ArgType{1,3} =` with **4** matches will produces **4** (event/relation) mentions each with **3** mentions for the argument key,`argName`.

## `{,e}`
- The argument must be matched **0** to `e` times (effectively optional).  If there are more than `e` matches for the argument pattern, a mention for each combination of matches of size `e` will be produced.

## `{s,}`
- The argument must be matched `s` or more times.  If there are fewer than `s` matches for the argument pattern, no mention will be produced.


odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
```diff
@@ -146,20 +174,44 @@ class ArgumentPattern(
...
+      // Kleene star
+      case (_, KleeneStar(true)) => Seq(matches)
+      case (_, KleeneStar(false)) => Nil
+      // One or more
+      case (_, OneOrMore(true)) => Seq(matches)
+      case (_, OneOrMore(false)) => matches.combinations(1).toList
+      // ranged quantifier w/ min and max reps
+      case (_, RangedQuantifier(Some(minRep), Some(maxRep))) =>
+        val reps = (Seq(matches.size) ++ (0 to maxRep)).filter(_ <= matches.size).max
+        matches.combinations(reps).toList
```
marcovzla: Consider a different way of writing this, for example, this may be more readable:
```
if (matches.size < minRep) Nil
else if (matches.size > maxRep) matches.combinations(maxRep).toList
else Seq(matches)
```

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
```diff
@@ -146,20 +174,44 @@ class ArgumentPattern(
...
     } yield (m, path.reverse) // paths were collected in reverse
-    if (matches.isEmpty) Nil
-    else if (size.isDefined) matches.combinations(size.get).toList
-    else Seq(matches)
+    (matches, quantifier) match {
+      // no matches
+      case (Nil, _) => Nil
+      // no quantifier w/ some matches
+      case (_, NullQuantifier) => matches.combinations(1).toList
+      case (_, OptionalQuantifier) => Seq(matches)
```
marcovzla: shouldn't this be `matches.combinations(1).toList`?

odin/src/main/scala/org/clulab/odin/impl/Quantifier.scala
```diff
@@ -0,0 +1,37 @@
+package org.clulab.odin.impl
+
+/**
+  * Represents a regex quantifier
+  */
+trait Quantifier {
```
marcovzla: Since this is specific to `ArgumentPattern`, maybe rename to `ArgumentQuantifier` or something similar?

odin/src/main/scala/org/clulab/odin/impl/Quantifier.scala
```diff
@@ -0,0 +1,37 @@
...
+  // whether or not the quantifier is greedy (true) or lazy (i.e., greedy = false)
+  val greedy: Boolean = true
```
marcovzla: Please remove `greedy`.

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
```diff
@@ -30,19 +30,47 @@ class DependencyPatternCompiler(unit: String, resources: OdinResourceManager) ex
...
+        //new ArgumentPattern(name, label, pat, required = true, quantifier = NullQuantifier)
+      // exact count
+      case name ~ ":" ~ label ~ Some(exact: Int) ~ "=" ~ pat =>
+        new ArgumentPattern(name, label, pat, required = true, quantifier = ExactQuantifier(exact))
+      // open range quantifier
+      // make combinations of the largest value found in the range (i.e., in {2,5} if 5 matches found, create combos of 5)
+      case name ~ ":" ~ label ~ Some( "{" ~ Some(minRep: Int) ~ "," ~ None ~ "}" ) ~ "=" ~ pat =>
+        new ArgumentPattern(name, label, pat, required = true, quantifier = RangedQuantifier(minRepeat = Some(minRep), maxRepeat = None))
+      case name ~ ":" ~ label ~ Some( "{" ~ None ~ "," ~ Some(maxRep: Int) ~ "}" ) ~ "=" ~ pat =>
+        new ArgumentPattern(name, label, pat, required = true, quantifier = RangedQuantifier(minRepeat = None, maxRepeat = Some(maxRep)))
```
marcovzla: I think this should be `required=false`

marcovzla:
Thanks @myedibleenso!
I think this is a nice feature to have.
I added some inline comments.

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
```diff
@@ -146,20 +174,44 @@ class ArgumentPattern(
...
+      // One or more
+      case (_, OneOrMore(true)) => Seq(matches)
+      case (_, OneOrMore(false)) => matches.combinations(1).toList
+      // ranged quantifier w/ min and max reps
+      case (_, RangedQuantifier(Some(minRep), Some(maxRep))) =>
+        val reps = (Seq(matches.size) ++ (0 to maxRep)).filter(_ <= matches.size).max
+        matches.combinations(reps).toList
+      // ranged quantifier w/ min reps
+      case (_, RangedQuantifier(Some(minRep), None)) =>
+        val reps = (minRep to matches.size).filter(_ <= matches.size).max
```
marcovzla: How about: `if (matches.size < minRep) Nil else Seq(matches)`

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
```diff
@@ -146,20 +174,44 @@ class ArgumentPattern(
...
+        matches.combinations(reps).toList
+      // ranged quantifier w/ max reps
+      case (_, RangedQuantifier(None, Some(maxRep))) =>
+        val reps = (0 to maxRep).filter(_ <= matches.size).max
```
marcovzla: Similarly here:
```
if (matches.size > maxRep) matches.combinations(maxRep).toList
else Seq(matches)
```

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
myedibleenso: Right.  Thanks!

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
myedibleenso: Yes, you're right.  Something like `argName:ArgType? = ` should never have a size greater than 1.

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
myedibleenso: Good suggestion, thanks.

odin/src/main/scala/org/clulab/odin/impl/DependencyPatternCompiler.scala
myedibleenso: Right.  No need for combinations.

odin/src/main/scala/org/clulab/odin/impl/Quantifier.scala
myedibleenso: Yeah, I had the same thought.  Will do.

marcovzla:
Thanks @myedibleenso! This looks really useful

marcovzla:
I think this is ready to merge

PR #155, opened 2017-06-27
Specify desired graph for Odin graph matcher 

### LINKED ISSUE #153: Add graph field to dependency rules
Thinking ahead to the eventual inclusion of SRL graphs and universal dependencies, We ought to have a way of  specifying which graph a rule should be applied to.  

We're currently producing `stanford-basic` dependency parses wherever we are using `stanford-collapsed` dependencies.  There are times where it is necessary to write a rule in terms of `stanford-basic`, instead of `stanford-collapsed` (the default).  For example, currently you can't use a preposition as an event trigger (ex. "due to"), because these nodes are disconnected from the `stanford-collapsed` graph.

## Proposed changes

1. Deprecate `type: dependency` in favor of `type: graph`
2. Add `graph:` as an optional rule field field
3. [Rename `Sentence.dependenciesByType` to `Sentence.graphs`](https://github.com/clulab/processors/blob/93f4e18bd899272aa9f65402241bcbc09292debe/main/src/main/scala/org/clulab/processors/Sentence.scala#L32)

## Example
```yaml
- name: "basic-deps-example"
   # type: dependency should also work this way
   type: graph
   graph: "stanford-basic"
   label: Event
   pattern: |
     trigger = due to
     # the rest of your rule here
```

When applied, the rule would check `Sentence.dependenciesByType` for the value corresponding to `graph` if `graph` is specified.  If the `graph` field is not included in a rule of `type: graph` (or `type: dependency`), the matcher would default to `Sentence.dependencies` (currently this points to "stanford-collapsed").

myedibleenso:
These changes resolve #153 by introducing `type: graph` and the new rule field, `graph:  "my-desired-graph-here"`.  This allows rules to be applied to Stanford basic dependencies, collapsed, SRL parses, and eventually universal dependencies.  Currently, the allowed values for `graph` are [limited to the following options](https://github.com/clulab/processors/blob/da7ec8a966db633cd889ac0292add8da34b4ba5f/odin/src/main/scala/org/clulab/odin/impl/OdinConfig.scala#L20).

The value of `graph` is also respected when checking a token's `incoming` or `outgoing` relations.

Note that in order to preserve the old behavior, the value of `graph: ` will be ignored whenever the type is specified as `dependency` (the new default is `graph`).

## Summary of Changes
- Renamed Odin classes containing Dep and Dependency to Graph
- Default rule `type` is now `graph`, instead of `dependency`
- The graph/parse to be used by a rule can now be selected via the `graph: ` field, which can be [set to one of several values](https://github.com/clulab/processors/blob/da7ec8a966db633cd889ac0292add8da34b4ba5f/odin/src/main/scala/org/clulab/odin/impl/OdinConfig.scala#L20).  Its default remains the same as always ("stanford-collapsed").
- Using `type: dependency` will ignore any specified value for `graph: `.
- Introduction of `OdinConfig` to simplify constructors
- Additional tests to cover new features

## Examples

The is token rule will use "stanford-basic" dependencies for its `incoming` and `outgoing` constraints
```yaml
- name: "date-token-basic-1"
  label: Date
  graph: "stanford-basic"
  type: token
  pattern: |
    [entity=DATE & incoming=pobj]
```

These graph-based rules will use basic Stanford dependencies.
```yaml
- name: "date-graph-basic-1"
  label: Event
  graph: "stanford-basic"
  pattern: |
    trigger = [lemma="disappear"]
    person: Person = nsubj
    location: Location = prep [lemma="in"] pobj nn?
    date: Date = prep [lemma="in"] pobj

- name: "date-graph-basic-2"
  label: Event
  type: graph
  graph: "stanford-basic"
  pattern: |
    trigger = [lemma="disappear"]
    person: Person = nsubj
    location: Location = prep [lemma="in"] pobj nn?
    date: Date = prep [lemma="in"] pobj
```


marcovzla:
I like having the ability to match over different graphs. I also like the introduction of OdinConfig.
Maybe add a comment explaining that OdinConfig is meant to be used in RuleReader, etc.
BTW, RuleReader is getting ugly, but that is left for some other PR

PR #157, opened 2017-07-11
Added CluProcessor: for now tokenization, lemmatization, POS tagging

MihaiSurdeanu:


.sbtopts
```diff
@@ -1,3 +1,3 @@
--J-Xmx5G
+-J-Xmx24G
```
myedibleenso: @MihaiSurdeanu, can we change this back to 5G?

MihaiSurdeanu:
Fixed in a separate PR. This was my bad.

PR #180, opened 2017-09-22
Bug fix: fixed model loading for the maltparser models, which were not loading properly from the processors-models jar.

MihaiSurdeanu:


myedibleenso:
🍾  🎊  

So this solution works when publishing locally and importing into another project?

MihaiSurdeanu:
Yes, it does :)

hickst:
Gus has a point: do we know that mkTmpDir will work when the library is used in a web application, which severly restricts filesystem access to a *root* filesystem (sandbox)?

PR #187, opened 2017-10-06
Redo uses of io.Source which confused compiler.

hickst:
Made the usages of `io.Source` consistent across many files by explicit import of `scala.io.Source`. This solves a problem I ran into where these usages suddenly confused the compiler when a single Akka library dependency was added.


MihaiSurdeanu:
@hickst: this is fine. Ok to merge.

PR #197, opened 2017-11-10
upgrade corenlp to the latest revision 3.8.0

dispalt:
Just some small changes to method signatures and objects.

### DIFF
--- a/corenlp/src/main/scala/org/clulab/processors/corenlp/CoreNLPProcessor.scala
+++ b/corenlp/src/main/scala/org/clulab/processors/corenlp/CoreNLPProcessor.scala
@@ -74,7 +74,7 @@ class CoreNLPProcessor(
       val stanfordTree = stanfordParse(sa)
 
       // store Stanford annotations; Stanford dependencies are created here!
-      ParserAnnotatorUtils.fillInParseAnnotations(false, true, gsf, sa, stanfordTree, GrammaticalStructure.Extras.NONE)
+      ParserAnnotatorUtils.fillInParseAnnotations(false, true, gsf, sa, util.Arrays.asList(stanfordTree), GrammaticalStructure.Extras.NONE)
 
       // save our own structures
       if (stanfordTree != null) {
@@ -123,7 +123,7 @@ class CoreNLPProcessor(
       } catch {
         case e: Exception =>
           System.err.println("WARNING: Parsing of sentence failed, possibly because of out of memory. " +
-            "Will ignore and continue: " + edu.stanford.nlp.ling.Sentence.listToString(words))
+            "Will ignore and continue: " + edu.stanford.nlp.ling.SentenceUtils.listToString(words))
       }
 
       //println("SYNTACTIC TREE: " + tree)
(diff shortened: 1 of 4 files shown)


corenlp/src/main/scala/org/clulab/processors/fastnlp/FastNLPProcessor.scala
```diff
@@ -61,8 +61,8 @@ class FastNLPProcessor(
       val gs = stanfordDepParser.predict(sa)
 
       // convert to Stanford's semantic graph representation
-      val basicDeps = SemanticGraphFactory.makeFromTree(gs, SemanticGraphFactory.Mode.BASIC, GrammaticalStructure.Extras.NONE, true, null)
-      val collapsedDeps = SemanticGraphFactory.makeFromTree(gs, SemanticGraphFactory.Mode.CCPROCESSED, GrammaticalStructure.Extras.NONE, true, null)
+      val basicDeps = SemanticGraphFactory.makeFromTree(gs, SemanticGraphFactory.Mode.BASIC, GrammaticalStructure.Extras.NONE, null)
+      val collapsedDeps = SemanticGraphFactory.makeFromTree(gs, SemanticGraphFactory.Mode.CCPROCESSED, GrammaticalStructure.Extras.NONE, null)
```
dispalt: The `threadSafe` parameter was removed.

corenlp/src/main/scala/org/clulab/processors/corenlp/CoreNLPProcessor.scala
```diff
@@ -74,7 +74,7 @@ class CoreNLPProcessor(
       val stanfordTree = stanfordParse(sa)
 
       // store Stanford annotations; Stanford dependencies are created here!
-      ParserAnnotatorUtils.fillInParseAnnotations(false, true, gsf, sa, stanfordTree, GrammaticalStructure.Extras.NONE)
+      ParserAnnotatorUtils.fillInParseAnnotations(false, true, gsf, sa, util.Arrays.asList(stanfordTree), GrammaticalStructure.Extras.NONE)
```
dispalt: I am not sure about this one.

MihaiSurdeanu: Seems like the right change.

MihaiSurdeanu:
Thanks Dan!
We are not quite ready for this change, because reach (our biomedical IE system) depends on Stanford dependencies, whereas this version of CoreNLP uses a different representation (universal dependencies). Let's keep this in a branch for a few weeks.

@hickst, @marcovzla, @myedibleenso: I would like to start a major version of processors (7.x) which uses the latest CoreNLP, and we do a complete move to universal dependencies. This means that Reach will have to stay with processors 6.x. Which further means that, if there are major changes to processors that would be needed in Reach (although I don't see any...), we would have to backport these changes to 6.x. Thoughts? Given the new WM project, I would like to do this asap.

hickst:
UD is being worked on. Closing this PR for now.

MihaiSurdeanu:
Yes, we hope to have the update to 3.8.0 up in a couple of days. It turns
out it was more complicated than expected due to several hidden changes in
CoreNLP.

PR #201, opened 2017-11-17
Add entityfinder

bsharpataz:
originally branched off of mentions_with_mods
- note: there is a small amt of cleanup I'd like to do sometime with the test files, but I need help (having trouble with class/package/whatever paths and class loading...

I have not yet transitioned to universal deps, that will be a later PR I think.


MihaiSurdeanu:
Generally good, but:

The grammars stored under ie/src/main/resources/org/clulab/grammar should be stored under a path that matches the corresponding packages in src/main/scala. That is, it should be ie/src/main/resources/org/clulab/ie/grammar.

I thought we're doing just entities in this PR? Why do you have events in ie/src/main/resources/org/clulab/grammar/taxonomy.yml? Are you doing causal events too? Shouldn't those be separated in different grammars + package?

Just checking: EntityConstraints.eventTriggers exists so we don't include event triggers in entities?

MihaiSurdeanu:
Also, I think the package names should o.c.openie.entities (not ie)

myedibleenso:
I think the `eventTriggers` check and corresponding avoid rules (+ var) should probably be removed here as well, as it is specific to one project at this point.  Removing it may involving tweaks to one or more tests, since these were ported from the influence search project.

myedibleenso:
If everyone agrees with the above, I can remove the event trigger stuff.

myedibleenso:
btw, why isn't this PR targeting `master`?

bsharpataz:
not targeting master bc the other project that depends on this (and will use a snapshot) also depends on the entity_with_state stuff.

you can remove event trigger stuff, or I can -- just let me know

myedibleenso:
>bc the other project that depends on this (and will use a snapshot) also

@bsharpataz, sorry, not following...isn't this pertinent to things beyond WM?  It doesn't depend on the stateful mentions stuff as it is, right?  

Anyway, I'll go ahead and remove the event trigger stuff.

myedibleenso:
I opened #202, which targets `master` and is limited to the openie/entity finder stuff.  Maybe we can close this one?

PR #202, opened 2017-11-17
OpenIE subproject with entity finder

myedibleenso:
## Summary of Changes
- `EntityFinder` trait and `RuleBasedEntityFinder` implementation
- tests for `RuleBasedEntityFinder`

**Note that the default rule set for the `RuleBasedEntityFinder` requires chunks.**


MihaiSurdeanu:
Looks good! One last comment: I think Transpared, StopWords, and StopPhrases should be loaded from resource files, rather than hard-coded.

myedibleenso:
Glad you brought that up.  Those are actually not needed here (they're used as part of of the deduplication procedure in the assembly component of influence search).  I guess they snuck in when we copied over the rest.

PR #204, opened 2017-11-22
Regression wrapper

danebell:
These changes create wrappers for `liblinear` and `libsvm`'s regression functions, analogously to the classifiers already present in `org.clulab.learning`. This required the creation of `RegDataset`, which is as similar as possible to `Dataset` except that labels are simply `Double`s rather than a free parameter `L`, and consequently no label `Lexicon`. In/out of `RegDataset`s from/to `SVMlight` format is available. 

Tests include artificial examples to guarantee identity with `liblinear` and `libsvm`'s results, as well as a real regression example from [this](https://archive.ics.uci.edu/ml/datasets/Communities+and+Crime) dataset.


MihaiSurdeanu:
This is great! Thanks!

MihaiSurdeanu:
Can you please add these new additions to CHANGES?

danebell:
Yes, definitely.

PR #209, opened 2017-11-27
Corenlp380

bsharpataz:
Switching to CoreNLP 3.8.0 with universal dependencies


MihaiSurdeanu:
@bsharpataz: I now see 3 model files for the dep RST parser:
[mihais@napoca ~/github/processors (corenlp380)]$ find . -name "model.dep.rst.gz"
./modelsmain/src/main/resources/org/clulab/discourse/rstparser/model.dep.rst.gz
./modelscorenlp/src/main/resources/org/clulab/discourse/rstparser/model.dep.rst.gz
./main/src/main/resources/org/clulab/discourse/rstparser/model.dep.rst.gz

And a similar story for the constituent parser...
There should be just one of each, both in this directory:
./modelscorenlp/src/main/resources/org/clulab/discourse/rstparser/

corenlp/src/main/scala/org/clulab/processors/corenlp/CoreNLPProcessor.scala
```diff
@@ -195,6 +205,7 @@ class CoreNLPProcessor(
     if(doc.sentences.head.syntacticTree.isEmpty)
       throw new RuntimeException("ERROR: you have to run the constituent parser before discourse parsing!")
 
+    // TODO Becky: uncomment these two lines after retraining
```
myedibleenso: ?

odin/src/main/scala/org/clulab/odin/impl/OdinConfig.scala
```diff
@@ -16,10 +16,10 @@ case class OdinConfig(
...
 object OdinConfig {
 
-  val DEFAULT_GRAPH = GraphMap.STANFORD_COLLAPSED
+  val DEFAULT_GRAPH = GraphMap.UNIVERSAL_ENHANCED
 
   def empty: OdinConfig = OdinConfig()
 
   // add newly registered graph types here
-  var VALID_GRAPHS = Set[String](GraphMap.STANFORD_COLLAPSED, GraphMap.STANFORD_BASIC, GraphMap.SEMANTIC_ROLES)
+  var VALID_GRAPHS = Set[String](GraphMap.UNIVERSAL_ENHANCED, GraphMap.UNIVERSAL_BASIC, GraphMap.SEMANTIC_ROLES)
```
myedibleenso: Please do not remove `GraphMap.STANFORD_COLLAPSED, GraphMap.STANFORD_BASIC` (they need to be added back to `GraphMap` as well).  Those are needed to apply rules to the old formats.

myedibleenso:
Please do not replace the old Odin tests.  Just add new ones.  Those test are important because we need to continue to be able to apply rules to the other dependency representations.

myedibleenso:
Would you mind also adding another odin test for a graph/dependency pattern containing a `:` (ex. `nsubj:pass`)?  I believe for the time being this will require quoting the string.

corenlp/src/main/scala/org/clulab/processors/corenlp/CoreNLPProcessor.scala
bsharpataz: @MihaiSurdeanu they're uncommented -- did you want me to keep them, comment them, or remove them?  for what it's worth, I didn't see these lines, and so I didn't comment them out for retraining -- do I need to repeat the training? (sorry if you told me about this -- I have no memory about it)

bsharpataz:
@myedibleenso re: the concern about `:` in deps (e.g. `nsubj:pass`), I added a test here:
  https://github.com/clulab/processors/blob/corenlp380/corenlp/src/test/scala/org/clulab/processors/TestCoreNLPProcessor.scala#L285
it passes, does this cover what you were concerned about?

bsharpataz:
@MihaiSurdeanu @myedibleenso about adding vs replacing the tests -- my (limited) understanding was that from here on out we were only supporting universal deps, and that 6.1.something would be used for reach, etc.  So, if that's the case, does this concern still hold?

myedibleenso:
@bsharpataz, yes, the concern still holds.

MihaiSurdeanu:
Can I please get a summary of what the issues are?

bsharpataz:
@ Mihai -- summary of the outstanding concerns in this "thread" or of that concern in particular?

MihaiSurdeanu:
Concerns about merging corenlp380 into master.

bsharpataz:
@MihaiSurdeanu I believe the outstanding concerns (mostly quoting comments from above):
- The removal of GraphMap.STANFORD_COLLAPSED, GraphMap.STANFORD_BASIC (they need to be added back to GraphMap as well). Those are needed to apply rules to the old formats.
- The removal of the old Odin tests vs just adding new ones. per @myedibleenso "Those test are important because we need to continue to be able to apply rules to the other dependency representations." I was wondering if that was true as 7.0 will move to universal deps only.  Maybe I am confused about which odin tests were removed.
- per me: Please clarify comment in line 208 of CoreNLPProcessor.scala (see my more detailed comment above)

MihaiSurdeanu:
> On Nov 28, 2017, at 1:02 PM, bsharpataz <notifications@github.com> wrote:
> 
> @MihaiSurdeanu I believe the outstanding concerns (mostly quoting comments from above):
> 
> 	• The removal of GraphMap.STANFORD_COLLAPSED, GraphMap.STANFORD_BASIC (they need to be added back to GraphMap as well). Those are needed to apply rules to the old formats.

I would prefer not to add STANFORD_* dependencies back, for two reasons:
- CoreNLP no longer produces them.
- We would have to maintain two different syntactic representations, which is mostly redundant.

> 	• The removal of the old Odin tests vs just adding new ones. per @myedibleenso "Those test are important because we need to continue to be able to apply rules to the other dependency representations." I was wondering if that was true as 7.0 will move to universal deps only. Maybe I am confused about which odin tests were removed.
> 	• per me: Please clarify comment in line 208 of CoreNLPProcessor.scala (see my more detailed comment above)
Please remove that comment. That is a left over that no longer applies.

myedibleenso:
>I would prefer not to add STANFORD_* dependencies back

Minimally, they should be added back to `OdinConfig`.  We're talking about two strings that allow us to match over those representations...

MihaiSurdeanu:
Sorry, what exactly are you saying? Just keep them in OdinConfig, even if they’re gone in GraphMap?

myedibleenso:
>@myedibleenso re: the concern about : in deps (e.g. nsubj:pass), I added a test here:
https://github.com/clulab/processors/blob/corenlp380/corenlp/src/test/scala/org/clulab/processors/TestCoreNLPProcessor.scala#L285
it passes, does this cover what you were concerned about?

No, I meant an Odin test for a graph/dep rule which matches over such edges.

myedibleenso:
>Sorry, what exactly are you saying? Just keep them in OdinConfig, even if they’re gone in GraphMap?

Yes, if you're fighting to remove them from `GraphMap`.  Those strings in OdinConfig define which graphs can be matched against (if present).

MihaiSurdeanu:
We can actually keep them in GraphMap. Just don’t populate them anymore (in CoreNLP*).

myedibleenso:
>Just don’t populate them anymore

Yes, I didn't expect them to be populated (I'm all for embracing UD as you know).  The inclusion of these strings is for matching over/loading in older datasets.

bsharpataz:
> No, I meant an Odin test for a graph/dep rule which matches over such edges.

Oh, OK, gotcha.  I removed the old rule and added a rule here: https://github.com/clulab/processors/blob/corenlp380/odin/src/test/scala/org/clulab/odin/TestGraphPattern.scala#L532

please let me know if more is needed.  You were right about the quotes (for rule compilation), also regex /s work too

myedibleenso:
>I removed the old rule and added a rule here

By "removed the old rule" do you mean you removed the test you had in `processors-corenlp`, or did I miss a commit somewhere?  

Sorry for the confusion (that test seemed useful as well).  In any case, thanks for adding the Odin test!

myedibleenso:
From my perspective, all that's left here is to add the strings back to `OdinConfig` and/or `GraphMap`, as well as the old `odin` tests (the ones that operated over the serialized doc using the SD parse).

bsharpataz:
Oh, haha, ok, I'll put it back and we can have both! thanks! :)

bsharpataz:
@myedibleenso I added back the older representations in GraphMap, OdinConfig, and Sentence.  Re: the changed tests, which tests are you specifically referring to?  I changed one bc a Gene_or_gene_sequence (or whatever) was being tested for, but it wasn't in either the corresponding lexicon or the overrides file.  OR are you referring to the tests in TestGraphPattern?

myedibleenso:
>OR are you referring to the tests in TestGraphPattern
Yes, those in `TestGraphPattern`.

myedibleenso:
For the record, 47f3e1d isn't necessary for odin to handle "legacy" graph formats.  This is how the graphs are retrieved: https://github.com/clulab/processors/blob/master/odin/src/main/scala/org/clulab/odin/impl/Graph.scala#L7-L13

bsharpataz:
@myedibleenso to add back the GraphPattern tests, seems I need to add `graph: "stanford-collapsed"` to all of them -- is this OK in the spirit of not changing old tests?  sorry to ask such a basic/dumb question

myedibleenso:
That's totally fine.  It makes sense that the default graph was changed given the switch to UD.

bsharpataz:
cool -- tests added back, is it OK how I separated them?  or do you want
them all in the same class?

myedibleenso:
>tests added back

Thanks!

>is it OK how I separated them? 

Sure, I guess.  So long as they run and pass!

myedibleenso:
Which they are: 
- https://travis-ci.org/clulab/processors/builds/309179402#L1648

bsharpataz:
woohoo! thanks gus!

bsharpataz:
hey -- where are we at with this?  If no more changes are needed (I think they're not), can someone try resetting the travis build? seems to have hung somewhere weird

MihaiSurdeanu:
- Has scala2.12 been merged in master?
- We decided to rename Modification in processors to something else, so it doesn’t conflict with Modification in Reach. Has this been done?

hickst:
I'm not sure who you are asking. Was someone given the task to do the renaming and make sure that it works with Reach? 

As per my emails a couple days ago, the Scala 2.12 branch was up-to-date as of last Friday but is awaiting the `Modification` issue to be merged to master (and thence to the 2.12 branch, then tested, then tested with Reach).

 > Has scala2.12 been merged in master? - We decided to rename Modification in processors to something else, so it doesn’t conflict with Modification in Reach. Has this been done?

MihaiSurdeanu:
- Let’s agree on a new name first. I think Marco had a good idea.
- Then, Tom, please rename it. Please note that this has to be back ported into 6.2.0 because master is now ahead of that and no longer compatible. Tom, please check that Reach unit tests pass.
- Then, Becky, please merge scala2.12 into master, and make sure processors unit tests pass.
- Then I will release.

myedibleenso:
@bsharpataz, looks like some conflicts have surfaced due to recent merges.

PR #214, opened 2018-02-02
adding ability to import variables

bsharpataz:
@marcovzla  added the ability to import vars, which will allow us to keep triggers/variables in one yml file to be references by others.  
as was done with taxonomy, a string following the vars: def will be treated as a path

(@marcovzla there were actually no additional changes on my local to push)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # Changes
++ **7.0.1** - Added feature to Odin to import variables from a file. Works similarly to taxonomies.
 + **7.0.0** - Updated library to use Scala 2.12 by default. Add ability to remove lexicon members.
 + **7.0.0** - Retrained our two discourse parsers (constituent- and dependency-syntax based) to work with universal dependencies.
 + **7.0.0** - Switched to CoreNLP 3.8.0 and universal dependencies.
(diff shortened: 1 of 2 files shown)


kwalcock:
This is nitpicky, but can you instead of making this change three times,
-    val vars = getVars(master)
+    val vars = master.get("vars").map(readVars).getOrElse(Map.empty[String, String])
just do something like this once in getVars?
-      .map(_.asInstanceOf[JMap[String, Any]].asScala.mapValues(s => cleanVar(s.toString)).toMap)
+      .map(readVars)
The - line could be commented out and the + line explained, if necessary.

marcovzla:
I agree with @kwalcock. It would be better to modify `getVars()` instead of repeating code.
I am currently travelling so I can't do it until later tonight, if you are willing to wait.

bsharpataz:
um... sure??? I can wait, this is SO not a becky thing (Marco did it for me
fast yesterday) and I'm nervous to make changes (which is maybe dumb), but
if you can do it, Mihai, I'd really appreciate it!

marcovzla:
I updated the CHANGES file

MihaiSurdeanu:
Ok. 7.0.1 has been released. It will take a few hours to show up on maven.

PR #231, opened 2018-03-18
Make sure paths can be serialized

kwalcock:


kwalcock:
This is not yet ready, but the request had to be made to activate Travis testing.  The complete tests works under IntelliJ and Eclipse, but not for sbt.  It complains about DirectedGraph not being accessible during deserialization.  I have yet to understand how that can be.

kwalcock:
This change is necessary for Mention serialization, which was broken.  No client seemed to have noticed before Eidos.

PR #232, opened 2018-03-19
Add test for equality and fix bug to make it pass

kwalcock:


kwalcock:
That is, equality of Mentions.  Can the recent modifications be made available to Eidos in a release sometime soon?

Issue #122, opened 2016-12-13
Cross build for Scala 2.11 and 2.12


cory-p-oncota:
Currently the library is unusable in Scala 2.12 due to the dependency on scala-parser-combinators.

The examples here demonstrate how to cross build for scala core modules:
https://github.com/scala/scala-module-dependency-sample

dispalt:
You mind if I take a shot at this?

MihaiSurdeanu:
Go for it.
Thanks!

hickst:
Closed by pull request #213

hickst:
Cross compilation does not work: Scala 2.11 ForkJoin code has been obsoleted in 2.12 in favor of Java ForkJoin code. This breaks cross-compilation.

Since processors has been updated to use Scala 2.12, I recommend that we do not maintain a separate 2.11 version unless we have demonstrable need for it. Recommend we change the status of this issue to `WON'T FIX` and re-close it.

myedibleenso:
@hickst, it looks like this is only an issue in the `learning` package.  Can we not cross-compile the other subprojects for 2.11?  Spark is incompatible with 2.12, so we are stuck using ancient releases.

MihaiSurdeanu:
Thanks @marcovzla!
I will release processors next.

PR #237, opened 2018-04-13
Cross-compilation for 2.11.x and 2.12.x

myedibleenso:
Many libraries that we're making using of in other projects are still stuck on 2.11.  This PR addresses #122 by reverting to the use of `scala.concurrent.forkjoin.ForkJoinPool` for thread-limited parallelization in these two places within the`learning` package:

- https://github.com/clulab/processors/blob/ef9bd61d23fde39707da313ed2cfa683b8a26dc6/main/src/main/scala/org/clulab/learning/Datasets.scala#L179
- https://github.com/clulab/processors/blob/ef9bd61d23fde39707da313ed2cfa683b8a26dc6/main/src/main/scala/org/clulab/learning/RFClassifier.scala#L97

This change produces a (seemingly innocuous) warning under  2.12, where use of `scala.concurrent.forkjoin.ForkJoinPool` is deprecated and aliased to `java.util.concurrent.ForkJoinPool`.   

It is my understanding that the implementations for `ForkJoinPool` differ in 2.11 and 2.12. (see https://github.com/akka/akka/issues/18262).  Based on the benchmarks posted in that `akka` issue, we might want to consider using the monix FJP instead:

- https://github.com/monix/monix-forkjoin

I also changed the travis configuration to test on 2.12.4 and 2.11.11.

You can publish both 2.11 and 2.12 versions locally with `sbt +publishLocal`.


myedibleenso:
Sorry, @kwalcock, that may have been an accident.  I would like to hear opinions about opting to instead switch to the `monix` FJP, though.  I hesitate to switch mostly because it feels like processors main/core already has too many dependencies.  

I kind of wish we could move `sequences` and `learning` (the FJP stuff is only here) into its own subproject, but it isn't easy to extricate at this point, as `CluProcessor` lives in main and depends on `sequences` (and `sequences` depends on `learning`).

Regarding the warnings in Eidos, those were fixed by #235.  Here is the [dependency graph](https://github.com/jrudolph/sbt-dependency-graph) for [`eidos`](clulab/eidos@e62f44e) :  

dependencies

These are the conflicts in `eidos` that I believe are generating warnings:

- `netty` version conflict originating from `processors` (`edu.washington.cs.knowitall.nlptools` vs. `akka`)

- `xml-apis` version conflict originating from `processors` (`ai.lum.common` vs `edu.washington.cs.knowitall.nlptools`)

marcovzla:
I made a few fixes, and made sure travis is actually running tests for both scala versions.
The deprecation warning @myedibleenso mentioned is still there, but I would suggest we keep it for now and we can add monix-forkjoin in a future PR if needed.

marcovzla:
closes #122

PR #239, opened 2018-04-30
Add hooks so that mention subclasses can be generated

kwalcock:


kwalcock:
This is the smallest set of changes I could think of to allow clients to create their own kinds of mentions.  It just refactors some methods so that a subclass can inherit and then override a minimal amount of functionality.  In other words, it's a hack, but one that preserves compatbility.  The better way would be to allow for use of a mention factory as an argument somewhere.  This has not yet been run with Eidos, so please don't merge yet.  This is more a request for comments.

marcovzla:
This looks good to me.
I would agree, however, that the factory approach would be better. It would allow us to create custom mentions without having to override so many things. If we have time, then I would prefer to explore that option.

PR #243, opened 2018-05-11
Check that array has head before calling it

kwalcock:


kwalcock:
CluProcessor has no more cases like this and in the meantime I remember checking.  However, most of the other processors do the same thing.  (This seems like copied code.)  That they don't have problems may be because some preprocessor of tokenizer happens to take care of the situation.  This seems more like luck than design, so I'd like to add the guard anyway.  It will be one less thing to worry about.

kwalcock:
That kind of check for proper pipeline execution is sprinkled throughout the processors.  Rather than double check sentence length everywhere, I propose having Processors.annotate(text) just not annotate(doc) when there are no sentences.  (That's the change in the most recent push.)  Since most methods are public, someone could still get around this and call something directly that results in an exception, but they have to try.  If all the checks need to be guarded, I think that a more formal pipeline check needs to be implemented.  If this small change passes Travis, I'm OK with leaving it at this minor adjustment.

PR #244, opened 2018-05-16
Added support for class weights in LibSvm's classifier

enoriega:
An overloaded variant of the train method of _LibSvmClassifier_ allows for optional weights to handle class balancing.

I had to add support for this for FocusedReading and believe it can be useful for the users in general


MihaiSurdeanu:
@enoriega: Good idea. But please see compilation errors in the Travis log.

enoriega:
@MihaiSurdeanu now it passes all the tests

PR #245, opened 2018-05-23
Perform lazy evaluation of hash code and store it

kwalcock:
A lazy val cannot be overridden, apparently, so hashCode is not lazy itself.  Its cachedValue is stored in cachedHashCode.  That in turn makes use of a method than can be overridden to calculate the value when necessary: calculateHashCode.

### DIFF
--- a/odin/src/main/scala/org/clulab/odin/Mention.scala
+++ b/odin/src/main/scala/org/clulab/odin/Mention.scala
@@ -165,7 +165,11 @@ trait Mention extends Equals with Ordered[Mention] with Serializable {
 
   def precedes(that: Mention): Boolean = this.compare(that) < 0
 
-  override def hashCode: Int = {
+  override def hashCode: Int = cachedHashCode
+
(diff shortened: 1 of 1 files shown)


kwalcock:
If this passes the processors tests, it will be deployed locally for reach and tested.  Please don't merge yet.  In addition, several reviewers will be tagged.

kwalcock:
This code seems to pass all the processors, reach, and eidos tests.

kwalcock:
I was going to ask whether I should merge this and then make a release.  I may have permissions from Sonatype to do a release, but not from GitHub to merge.  Perhaps, however, I should leave Processors to someone else anyway and just do those for Eidos.  When this one is eventually available, I can complete the Eidos changes so that we're not reading one document per line on files that come from PDF conversions.

kwalcock:
It looks like I have some new permissions, so I'll be figuring out how to release this.  The change log will be updated, release incremented, a tag created, etc.

PR #252, opened 2018-06-13
Replace more unicode characters

kwalcock:
Uses spaces if need be.


kwalcock:
Is the idea then that Eidos make use of these conversion functions?  Our \u2022 will turn into * and \u27a4 will be -> because these are known Unicode values.

MihaiSurdeanu:
Yes, exactly.

PR #253, opened 2018-06-18
Increase memory in .sbtopts

kwalcock:
The Travis build is periodically failing with out of memory complaints.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,3 +1,3 @@
--J-Xmx5G
+-J-Xmx6G
 -J-XX:+CMSClassUnloadingEnabled
 -J-XX:+UseConcMarkSweepGC


MihaiSurdeanu:
I'm OK with this. But Travis still fails...

kwalcock:
I see that the builds still failed with memory problems and wouldn't merge this one yet.  An investigation has been launched.
```
06:45:28.572 [pool-5-thread-6-ScalaTest-running-TestLibSVMClassifier] DEBUG org.clulab.learning.LibSVMClassifier - Using 2000 datums.
Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x000000075cb45000, 1665904640, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 1665904640 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /home/travis/build/clulab/processors/hs_err_pid3746.log
The command "sbt ++$TRAVIS_SCALA_VERSION 'test-only -- -l NeedsExternalBinary'" exited with 1.
Done. Your build exited with 1.
```

kwalcock:
Some people have pointed out a possible memory leak in older versions of sbt including the substantially old one (0.13.13) that is used here.  Is there a particular reason we're using that (other than what I would run into by trying it out myself)?

MihaiSurdeanu:
No, we can definitely upgrade sbt!

bsharpataz:
I think we may have to use newer versions of some play stuff (for the webapp) -- @marcovzla made these changes for clay when using odin for a new project -- update our files in the project/ dir to match the ones here: https://github.com/ml4ai/musica_odin/tree/master/project

kwalcock:
@bsharpataz, could you check that link?  Or maybe it's a page I don't have access to.

The tests passed at 5.5GB, which might have left enough room for whatever else was next in the tests.  I'll update the sbt version number in Eidos first, because that's where I can hack on it more, and make sure it doesn't cause problems.

bsharpataz:
sorry -- content:
change project/build.properties to:
`sbt.version=1.1.6`
and change project/plugins.sbt to:
```
addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.6.6")
addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.8")
addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "5.2.4")
addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0")
addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "2.3")```

marcovzla:
+1 to upgrading sbt.

However, we may not have enough memory to allocate in the travis env.
I believe we currently have 7.5GB, but that includes OS, etc.
https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System

So maybe we should look at how to reduce memory consumption instead.

kwalcock:
Reducing memory (and maybe processor?) consumption sounds good, but I hope to postpone that for a different PR.  I'm modernizing the sbt and plugins on eidos shortly to match your settings, @marcovzla.  They'll get back to processors at some point.

PR #254, opened 2018-07-10
Repetitions

marcovzla:
Allow repetitions in dependency patterns to start from zero explicitly.

`range{0,5}` should be equal to `range{,5}`


marcovzla:
@bsharpataz fyi

PR #266, opened 2018-08-09
normalize unicode text

bsharpataz:
This is Marco. I am adding code to normalize unicode text. Also changed the code to preserve accents, to be used with the Portuguese parser.


MihaiSurdeanu:
@gcgbarbosa: the tests for this PR fail. Can you please look into this?

MihaiSurdeanu:
@gcgbarbosa: is this chunking model trained with the option "-bi 10"? I think that will improve robustness on this small(ish) dataset, as it is a bidirectional model.

gcgbarbosa:
Yep. I guess these are the same errors @kwalcock pointed before. From my understanding, they are due to unicode issues as you can see from this line:
[info] "L[á]" was not equal to "L[á]" (TestPortugueseCluProcessor.scala:21)

I guess @marcovzla already solved them.

marcovzla:
I still need to add a unit test for the issue that started all this, which is that in the sentence `O deputado João caçou Júnior.` the word `João` gets tokenized as `Jo a o`

marcovzla:
@MihaiSurdeanu tests pass now

PR #257, opened 2018-07-25
Portuguese and Spanish processor improvements

danebell:
All Portuguese and Spanish part-of-speech and parsing models have improved training and testing:

- Contractions are removed from training files.
- Separate tests are administered per test corpus.
- Part-of-speech models are BiMEMM
- Tests of SpanishCluProcessor have been added.

Additionally, Spanish tokenization and sentence splitting have been implemented:

- Spanish- (and Catalan-)specific characters have been added to the word forms, and are protected from normalization to non-accented equivalents.
- Common Spanish abbreviations and sentence-starts are enumerated to improve sentence splitting.
- Spanish contractions `del` and `al` are each split into their constituents

PR #267, opened 2018-08-16
travis cache

marcovzla:
Trying out travis cache as suggested in #257

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,4 @@ cache:
     - $HOME/.sbt
 script:
   - sbt ++$TRAVIS_SCALA_VERSION 'test-only -- -l NeedsExternalBinary'
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,5 +8,15 @@ scala:
 branches:
   only:
     - master
+before_cache:
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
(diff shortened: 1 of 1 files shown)


kwalcock:
P.S.  If it helps at all, the effect probably won't be seen until the second build.  The first build will likely be slower as it has to fetch all the dependencies as normal plus write them all back out to the cache.

MihaiSurdeanu:
I'd like to see performance differences on this. I'll play with this.
Should we merge this?

kwalcock:
You'll probably want to build/test again.  Click "Show all checks" below and have Travis perform the same thing again and check the time.

MihaiSurdeanu:
The first-time test completed in ~59 mins.

build1

Now running the second one.

MihaiSurdeanu:
I don't think this is working. The second test ran in the same time:

screen shot 2018-08-17 at 7 40 23 am

MihaiSurdeanu:
Well, maybe it downloaded all the dependencies in the 1:30 min of difference? Possible.
In any case, this is not a major speedup.

kwalcock:
It does seem like it used the cache, but then to less effect than desired.  I guess the mileage varied.  It seemed worth a try.

MihaiSurdeanu:
We could keep it... It helps a bit, and it doesn't hurt.

marcovzla:
Closing this since it hasn't been touched in a while.
Feel free to reopen.

PR #268, opened 2018-08-25
Dynet

kwalcock:
This PR is only here to see what Travis does with dynet.  Please do not merge.


MihaiSurdeanu:
@kwalcock: very elegant solution!

One question: how do we add the dependency to dynet into the "main" subproject?

MihaiSurdeanu:
I wonder if a simpler solution would be to create a completely separate project whose purpose is simply to store these fat jars. We can release it on maven separately, and simply include dependencies to it in processors...

kwalcock:
That strategy is looking better to me after my attempts to trick sbt into the desired behavior failed yesterday.  That's probably more my fault than sbt's.  There might need to be two GitHub projects, dynet11 and dynet12.  They would both be known to maven as dynet, but one would result in dynet_2.11 and the other dynet_2.12.  The current strategy results in some files like dynet_2.11_2.11.jar which may cause problems.

kwalcock:
This PR is outdated, so I will remove it.  There is a separate fatdynet repo for this code.

PR #269, opened 2018-09-05
New PT models

gcgbarbosa:
Re-trained models for chunking

- Normalised dataset
- Added re-trained model with split PROPNs
- Added new test file for new model

Re-trained models for parsing

- Normalised dataset
- Create new pos tags for verbs, based on PALAVRAS tagset
- Trained new model with the previous mod
- Added tests for new model

Re-trained pos model

- Re-trained pos model with the same dataset used to train parser model
- Added tests for new model


MihaiSurdeanu:
@gcgbarbosa: These are fine, but why are the performance values for the parser so low in TestPortugueseCluParser? Wasn't the performance of the dependency parser in the 80s now?

MihaiSurdeanu:
@gcgbarbosa: please see my question above.

gcgbarbosa:
@MihaiSurdeanu 

You are right. The metrics are above .8, but only for bosque test set (as we did not use gsd and pud as training data).

I am updating the TestPortugueseCluParser with the real metrics we are getting.

MihaiSurdeanu:
Ok. Ready to merge then @gcgbarbosa?

Then can we release processors, @kwalcock, @marcovzla ?

gcgbarbosa:
@MihaiSurdeanu 

I have done a little more testing on Eidos using the new models and it seems ok for me.

PR #271, opened 2018-09-07
colon and dash

marcovzla:
This PR changes the syntax of odin's stringmatchers so that they may now include colons and dashes without being quoted. This is useful for matching universal dependencies, e.g., writing `>nmod:poss` instead of `>"nmod:poss"`. Also for writing `[chunk=B-NP]` instead of `[chunk="B-NP"]`.


MihaiSurdeanu:
@marcovzla: this looks good to me. Just to make sure, can you please run the Reach unit tests with this? If they pass, I recommended merging and releasing.

marcovzla:
@MihaiSurdeanu all the tests on eidos pass. I'm not sure how to try this on reach, because a lot of tests would break just by switching to processors 7.2.
Is there a branch I should be using?

MihaiSurdeanu:
You’re right. We can try this on Reach yet. 
If Eidos passed, I think we’re fine. 
Merge?

PR #275, opened 2018-09-25
diaeresis support in portuguese tokenizer

marcovzla:
Added diaeresis support in portuguese tokenizer.

Note that the java file was generated with antlr 4.5.2 but the previous one was generated with antlr 4.6, so maybe someone wants to regenerate it with a newer antlr.


marcovzla:
there seems to be something wrong with the java code generated by antlr 4.7.1

```
[error] /home/travis/build/clulab/processors/main/src/main/java/org/clulab/processors/clu/tokenizer/OpenDomainPortugueseLexer.java:100: method does not override or implement a method from a supertype
[error] 	@Override
```

MihaiSurdeanu:
Maybe use 4.6?

PR #279, opened 2018-10-05
Sbt upgrade

myedibleenso:
## Summary of changes
- Upgraded `sbt` to next major version (`v0.13.x` -> `v1.x`)
- Support GPG v2 for release signing


MihaiSurdeanu:
Looks good, but tests are failing...

kwalcock:
These tests ended with out of memory errors.

mslinn:
It would be so nice to merge this PR.

MihaiSurdeanu:
All these make sense, but we still have the OOM error. Maybe reverting back to -J-Xmx5500M in .sbtopts would help? @marcovzla, what do you think?

mslinn:
6GB RAM is a tiny amount for this type of application.

marcovzla:
I agree it would be nice to merge this PR. Upgrading sbt would simplify things for me.
We haven't done so because Travis runs out of memory when executing the unit tests.
See https://travis-ci.org/clulab/processors/jobs/499513609#L2673
We wanted to figure out a way to decrease the memory consumption so that we can still use Travis, but no one is working on it as I know.

mslinn:
Seems like this PR is simple and straightforward. It works. If Travis is getting in the way, why not stop using it? Otherwise this project remains frozen in time, which would be a shame.

MihaiSurdeanu:
@marcovzla: does this PR pass if you set -J-Xmx5500M?

marcovzla:
It doesn't. I can take a look again next week.

MihaiSurdeanu:
@marcovzla, @myedibleenso, @mslinn: This was merged in master through a different branch. Our unit tests have been switched from Travis to Jenkins, where we have more RAM.

PR #286, opened 2018-11-19
Better handling of parentheses

MihaiSurdeanu:


kwalcock:
Does this address at all https://github.com/clulab/eidos/issues/300 ?  That code is from EntityConstraints, so these changes are probably at a lower level I guess.

Issue #294, opened 2019-01-18
Hashing of sentences takes up too much time


kwalcock:
Here's an incomplete record of emails regarding this issue which Becky and Marco discovered:

![image](https://user-images.githubusercontent.com/8679738/51404743-95046d00-1b11-11e9-8357-0fa0decdc1bd.png)

Yeah, those vars need to be made private, and accessed through setter methods, who will also reset the computed hash value. 
Should not be too hard to change… More annoying than hard, I think.

> On Jan 3, 2019, at 10:09 AM, Keith Alcock <keith@keithalcock.com> wrote:
> 
> Yes, probably, unless it is too problematic because of all the vars.
> 
> 
> On Thu, Jan 3, 2019 at 10:41 AM Mihai Surdeanu <surdeanu@gmail.com> wrote:
> I see… But Sentence hashing still matters, even if (optimally) just for the sentences that contain mentions. 
> So it’s probably still worth it to implement a caching mechanism for the hash value, no?
> 
> > On Jan 2, 2019, at 11:00 PM, Keith Alcock <keith@keithalcock.com> wrote:
> > 
> > I would think that most of the hashing that occurs with Eidos is just to make sure that Mentions are not used twice.  Mention.calculateHashCode does a complete hash of the document, including all sentences
> > 
> > val h4 = mix(h3, document.equivalenceHash)
> > 
> > even when we know there is only a single document in question.  When this is the case, probably a plain document.hashCode (or nothing at all) would do.  I can imagine it being difficult to separate the two cases cleanly, though.
> > 
> > 
> > 
> > On Sat, Dec 29, 2018 at 1:11 PM Rebecca <bsharp@email.arizona.edu> wrote:
> > While the synchronized NN is the main slowdown pretty much for sure, we did notice something else that may really speed things up if we can address it. 
> > 
> > Sentence equivalence hash is a def, not a val, so gets computed over and over.  This actually seems to account for at least 20-30% of the actual runtime (we didn't keep expanding):
> > 
> > 
> > 
> > maybe one solution is making it a val, a lazy val, or a var. but we don't know.  Either way, it is a lot of time spent computing hashes and maybe there is an opportunity for optimization.
> > 
> > Thanks guys!
> > Becky and Marco
> 
>

myedibleenso:
[...]

PR #295, opened 2019-01-19
Test the ambivalenceHash

kwalcock:
Please do not merge yet.  This PR is so that Travis runs the tests which do not work on my computer.

### DIFF
--- a/main/src/main/scala/org/clulab/processors/Document.scala
+++ b/main/src/main/scala/org/clulab/processors/Document.scala
@@ -52,6 +52,12 @@ class Document(val sentences: Array[Sentence]) extends Serializable {
     val h1 = mix(h0, sentencesHash)
     finalizeHash(h1, 1)
   }
+
+  def ambivalenceHash: Int = {
+    val h0 = stringHash(getClass.getName)
+    val h1 = mix(h0, orderedHash(sentences.map(_.ambivalenceHash)))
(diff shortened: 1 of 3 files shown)


kwalcock:
This change is discussed in issue #294.  It has been unit and regression tested in Eidos and unit tested in Processors.  equivalenceHash remains, but there is a new ambivalenceHash in Document and Sentence that is used to calculate the hashCode of a Mention.

MihaiSurdeanu:
This is fine, but I wonder if we can make it even faster. "words" are a deterministic transformation of "raw", so they are strictly not needed in the hash, if you have "raw" in there...

kwalcock:
Yes, that can be updated.  I was on the fence about it.  The transformation may be subject to change, but probably not on the timescale of sentence comparisons.  It will probably not make a lot of difference in performance which is probably more the result of caching.  PR momentarily.

kwalcock:
I'm seeing a complaint on an Eidos serialization test.  Please don't merge yet.

kwalcock:
That was my bad.  By using this.getClass.getName I was ruling out subclass objects from having the same hash.  Eidos needs to compare an EidosDocument to a Document.  However, rather than using a fixed stringCode as in the other cases, I went with Document.getClass.getName so that refactoring will work.  It should be OK now when Travis completes.

PR #298, opened 2019-01-24
Added support for OpenIE extractions into Processors

enoriega:
I hacked in the code to support OpenIE within processors. The data abstraction for this is a new field in the `Sentence` class that holds an _Optional_ array with relations extracted by OpenIE. 

Each relation is represented as an instance of `RelationTriple`(See https://github.com/clulab/processors/blob/openie_integration/main/src/main/scala/org/clulab/processors/RelationTriple.scala)

Here's an example of how it could be used:

```scala
val text = "Obama was born in Hawaii. He is our president."
val processor = new FastNLPProcessor(withRelationExtraction = true)
val doc = processor.annotate(text)

sentences.foreach{
  implicit s =>
    for(triple <- s.relations.get) { 
        println(s"Subject: ${triple.subjectText}")
        println(s"Relation: ${triple.relationText}")
        println(s"Object: ${triple.objectText}") 
        println(s"Confidence: ${triple.confidence}") 
    }
}
```
This feature is supported in`FastNLPProcessor` and `CoreNLPProcessor`. The annotations are optional and off by default, to avoid any performance penalty on existing code bases. The `Bio` variants of these processors don't support OpenIE as they're domain specific and most likely to be processed by REACH.

Finally, I also added support for serialization of these new annotations via `DocumentSerializer`.

There is a unit test for each processor, although it can be improved.


MihaiSurdeanu:
Thanks! I'll review this sometime today.

enoriega:
@MihaiSurdeanu I'm going to do a small improvement on the code and commit it. Please wait until it's up to approve it if applicable. You can still take a look to the architecture of the feature

myedibleenso:
- https://github.com/clulab/processors/pull/298/files#diff-8594a2bbe23f7328d006a690b5ef8bc7R330
- https://github.com/clulab/processors/pull/298/files#diff-69f12216d9aff01f4c020dcb73c9bde8R45
- https://github.com/clulab/processors/pull/298/files#diff-69f12216d9aff01f4c020dcb73c9bde8R190

etc.

I'm suggesting that instead the method __return__ some kind of `Seq[RelationTriple]`, rather than store it in the `Sentence`.

We don't store `Mentions` this way, so at least to me it's unclear why we'd want to store these triples within the Dcoument/Sentence.  I guess I was envisioning this much as the flow of information we see with Odin: `Document/Sentence -> Seq[RelationTriple]`.  

I see the token attributes (POS tags, lemma forms, etc.) and parse as more foundational information in comparison to the triples.  Does that make sense?  I'm curious to hear what others think, though.  Perhaps I'm in the minority on this : )

myedibleenso:
For the CoreNLP implementation, a converter utility like this could work to provide you with the information needed to return the triples: 

- https://github.com/clulab/processors/blob/ac6775e6cee6228f33cbcd00a51382f01a7bc306/corenlp/src/main/scala/org/clulab/processors/corenlp/CoreNLPUtils.scala#L110

If we went with that approach, I think we'd also need a `DirectedGraph` to `SemanticGraph` conversion, though.

bsharpataz:
FWIW I agree with Gus -- Triples seems much more analogous to Mentions from
Odin

enoriega:
Fair point. Still, we need to store it somewhere, as it is the result of a corenlp annotation executed when calling `annotate(doc)` and not computed in the fly.

Possibly the document object is a better home for the relation triples, akin to the discourse parse.

So far I emulated the lemmas, tokens, etc.

Would this be a cleaner solution?

MihaiSurdeanu:
I don't really see what's bad here. @enoriega is following the standard approach we have for these optional annotations, by storing them as Options...
We can, and should, exclude them from the hash function. But I don't see modifying Sentence in place as too bad, given the way it was designed.

myedibleenso:
Just to be clear: I think Open IE-style triples are useful and a welcome addition to processors. 

Anyway, it sounds like the opinions so far are of two camps.  It's Mihai's project, though, so it's ultimately up to him.  My reasoning is that these are closer to `odin.Mention`s and I do not think that mentions should _ever_ be stored in the doc or sentence.  I consider these triples (and Mentions for that matter) to be less standard/foundational that say part of speech tags and syntactic parses.

>we need to store it somewhere

My point was you really don't (`odin.Mention`s is an example of alternative way of thinking about this).  I'm suggesting that the Open IE extraction _not_ be a part of `annotate`, but instead be a separate call (hence the link to the converter).

I am of the opinion that `Sentence` is at risk of becoming too bloated (others might say "enriched").  Yes, I realize these annotations are all `Option`s.  I guess I just think this approach is untenable in the long term.  Following this trend, the next addition to `Sentence` could very well end up being `var sentenceEmbedding = Option[Embedding]`.

MihaiSurdeanu:
Well, I see these triples closer to semantic role frames, which I would store in the Sentence class...

myedibleenso:
Just a thought, but what about the triples in the form of a `DirectedGraph`?  One advantage to that approach is that we could also run matchers using Odin which could easily build upon this information.

MihaiSurdeanu:
^ this is a good idea!

enoriega:
Yes, this is also coming very soon.

myedibleenso:
@enoriega, if you do that (add it as a graph), then it might make sense to just add a utility to go from graph to `RelationTriple`.  Then you wouldn't need to store it redundantly in `Sentence` as `Option[Array[RelationTriple]`.

myedibleenso:
Here is where the graphs for the graph-based matcher are registered in `Odin`: 
- https://github.com/clulab/processors/blob/4890c0c64c9d107666330a1d8e3ca7371edc8729/odin/src/main/scala/org/clulab/odin/impl/OdinConfig.scala#L23-L30

enoriega:
@MihaiSurdeanu This should be good for what I meant to include in this pull request. Please let me know if I should merge

PR #300, opened 2019-01-28
Make HashTrie and LexiconNER better

kwalcock:
Do not merge this, please.  It is still in the test phrase.


kwalcock:
It as just for testing

PR #304, opened 2019-02-06
Save and load the RNN model

kwalcock:
This is an updated version which does not use fatdynet to save or load.  The model can be saved, loaded, and resaved resulting in a bitwise identical file.  Evaluation runs across the dev training sentences result in matching values.  Some test code remains so that runs can be made in reasonable times, so do look at the changes.  Some code was rearranged so that the RNNParameters could be set either by training or by loading.

### DIFF
--- a/main/src/main/scala/org/clulab/sequences/RNN.scala
+++ b/main/src/main/scala/org/clulab/sequences/RNN.scala
@@ -42,7 +42,10 @@ class RNN {
       sentences = MathUtils.randomize(sentences, rand)
 
       logger.info(s"Started epoch $epoch.")
-      for(sentence <- sentences) {
+      for (index <- 0 to 1000) {
+        val sentence = sentences(index)
+//      }
+//      for(sentence <- sentences) {
         sentCount += 1
         //logger.debug(s"Predicting sentence $sentCount: " + sentence.map(_.get(0)).mkString(", "))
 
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
Thanks @kwalcock!
This looks good. But can you please merge the dynet-rnn branch into this again? There were some changes there since this branch was created.

kwalcock:
I *think* that I beat you to it.  This PR is for kwalcock-rnn2 which is only one push ahead of dynet-rnn and no pull behind.  The old kwalcock-rnn was closed without a merge.

MihaiSurdeanu:
Cool. Then, merge?

kwalcock:
For testing in a reasonable time I commented out a few lines.  I'm not sure whether you want them changed back or not before a merge.  Otherwise it is OK.

MihaiSurdeanu:
That's fine. Should I merge then in mybranch, so I can play with it, extend it, etc?

kwalcock:
Sounds good.

MihaiSurdeanu:
But is this PR for merge into *master*? Then, I think we should wait a bit. 
I am merging it into dynet-rnn for now.

kwalcock:
Just to your branch, the work in progress, as it seemed to have happened.  Thanks.

PR #306, opened 2019-02-14
Load the x2i files

kwalcock:


kwalcock:
This may be an answer to

```
In this line:
https://github.com/clulab/processors/blob/dynet-rnn/main/src/main/scala/org/clulab/sequences/RNN.scala#L516

we need to be able to load() without access to to trainSentences, so users of our library don’t need to cary the training corpus with them, just to use our code. This means that we need to save two files: your rnn.dat, and another one that contains w2i, t2i, and c2i. I can do this. But is there a better solution?
```
The processors code probably has a library function for much of this, but here is a quick and dirty proof of concept.

MihaiSurdeanu:
Almost... 
To simplify and generalize the save/load process, I would prefer to have only 2 files on disk:
- The file of dynet parameters, which you already have.
- A second file that contains everything else that is custom to us, e.g., all x2i indexes. So, 1 file here, instead of 3.

kwalcock:
Ah, that notification hadn't arrived yet.  I'll put them in one file.

PR #308, opened 2019-02-17
Show how it would work with new version of fatdynet

kwalcock:
Please do not merge.  This shows how RNN.scala would work with the new version of fatdynet.  There is a PR for that at https://github.com/clulab/fatdynet/pull/8.  Imagine that load(dynetFilename:String, x2iFilename: String) was working with a model trained and saved from Python and couldn't reuse the Scala code.


kwalcock:
This was just an example, so I'm going to hide it now.  The branch is still available for reference.

Issue #302, opened 2019-02-04
LexiconNER (and HashTrie) needs to be smaller, faster


kwalcock:
Mihai wrote this once upon a time:

Also, on a different topic: processors has gotten pretty fat. Some of it is out of our control (CoreNLP models), but some of it is our stuff. In particular, we use a lot the LexiconNER class, which stores its internal data (basically multi-word dictionaries) as a HashTrie:

https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/struct/HashTrie.scala

This is a class that I implemented a while ago, and I am fairly sure it’s sub-optimal. At some point I’d like to brainstorm with you about optimizing it…

kwalcock:
Here is some of the discussion that followed, simply pasted from the email.

Hi Keith,
Sorry about the delay on this! Comments inline.

On Tue, Dec 11, 2018 at 1:09 PM Keith Alcock <keith@keithalcock.com> wrote:
So I've made lots of changes to HashTrie and LexiconNER which should not alter any results and should probably not have impact on other files or projects (only probably, because some public defs were moved).  I still have to do the LexiconNER object and verify with space and time measurements and do regression tests.  However, there are many other possible changes that may be worthwhile but will have consequences outside this environment (including serialization) and someone else should decide their fate.

1. uniqueStrings.  These are kept in HashTrie for statistics but probably at a significant price.  "Improvements" might be

- at least do a clear() after statistics are complete
- include a debug argument in constructor to disable statistics
- use a subclass to keep statistics only when necessary (e.g., in TestHashTrie)
- manage an Option[HashSet] from outside the class and usually use None as an argument to add(tokens) or clear as necessary

If they are not used in a significant way outside of the class (can't check right now, I don't have a laptop with IntelliJ here) then I agree they should at least be made optional. I think the simplest option would be the second one: include a debug argument in constructor, and have it set by default to false. And uniqueStrings should be an Option, as you say.
 

2. The labelling in HashTrie.findNormalized and LexiconNER.findLongestMatch.

This is done with string addition which may result in many instances of strings that are the same.  I've used some lazy vals and calculate only once per call in HashTrie or once per hit in LexiconNER, but this is still inefficient.  Here is one work around:

- Set the labels in the constructor to HashTrie (or special subclass) and create the B- and I- versions just once per KB.
- Let LexiconNer access these values so that this "val matchers: Array[(String, HashTrie)]" is not necessary.  The HashTrie will have this string/label.

You mean in places like this:
[...]

PR #311, opened 2019-02-26
Speed up LexiconNER

kwalcock:
This PR is meant to provoke Travis into running unit tests.  Please do not merge yet.  It will be tested with reach next.


MihaiSurdeanu:
Do unit tests in Reach pass with this change?
Also, would it be possible to get times for "sbt main/test" in Reach before/after this change

kwalcock:
It doesn't work with reach yet.  In one case reach skips the LexiconNER and goes directly for a HashTrie which I had eliminated.  I'm starting to put that back in so that reach builds without change.

FWIW there seems to be an existing compilation problem in HomeController that neither the build nor tests catch.

kwalcock:
main/test running on processors 7.4.4 ran in 5:04 minutes just now with the original code, and 7.4.5-SNAPSHOT ran in 3:27 minutes with the new code.  I wasn't expecting that much of a difference.  Some more numbers will appear at #302.

MihaiSurdeanu:
Nice!

I wonder what these numbers would look in Reach, assuming the same data structures are used?

kwalcock:
I wasn't clear above.  The times are from reach's main/test when build.sbt is configured to use the old and new version of processors.  reach seems to use the BioCluProcessor which in turn uses the LexiconNER which has been rewritten.  The BioCluProcessor is what I have been timing in processors as well via a new test case.  The intention is that it is ready to go and reach will be faster after the merge and republishing of processors and then an update to the dependency in reach.  There are a couple of other PRs ready for processors that should also go in.

kwalcock:
If someone needs/wants to test how this might affect some application of reach on some corpus, then doing a publishLocal on this branch (kwalcock-DuoHashTrie) and then changing reach's main/build.sbt procVer to 7.4.5-SNAPSHOT should do the trick.

MihaiSurdeanu:
Thanks! Last question: if you run runAllTests.sh in reach, do all tests pass?

kwalcock:
Yes, they do.

MihaiSurdeanu:
Cool. Then merge?

kwalcock:
If you please.  #295 and #305 should also be merged before the next release.  I'm not sure about the others.

MihaiSurdeanu:
Done. Can you please update the CHANGES file (let’s use as next version 7.5.0), and release?

PR #313, opened 2019-02-27
fixed bug in lazy quantifiers

marcovzla:


MihaiSurdeanu:
Can you please also update CHANGES for 7.5.0?

MihaiSurdeanu:
Ok to release now?

marcovzla:
i think so
@myedibleenso ?

myedibleenso:
Yes.  Would be nice to have even more tests, but at least we have a one : )

MihaiSurdeanu:
@kwalcock: can you please release? You have the permissions, right?

kwalcock:
I believe so.  Will wait for that one last approved merge.

PR #315, opened 2019-02-28
Use ordered hash

myedibleenso:
@kwalcock observed that `Sentence.equivalenceHash` was using an unordered hash for fingerprinting token attributes.  I suspect this could cause problems (imagine, for example, the same text annotated by two different processors, each of which set the same number of optional attributes to `None`, but differ in which attributes were set to `None`).  The `Sentence.ambivalenceHash` is unaffected by this change.


MihaiSurdeanu:
lgtm

PR #317, opened 2019-03-06
Fix calculateHashCode to use general trait rather than class name

kwalcock:
If class names are compared instead, reach crashes in MentionFilter.synPathContainsTrigger.  It is getting an unexpected mention there which has a synPath of None.  I'm not sure whether it should just be more careful there or whether the hash code is solely to blame.  This fixes the breaking problem, though.

### DIFF
--- a/odin/src/main/scala/org/clulab/odin/Mention.scala
+++ b/odin/src/main/scala/org/clulab/odin/Mention.scala
@@ -170,7 +170,7 @@ trait Mention extends Equals with Ordered[Mention] with Serializable {
   protected lazy val cachedHashCode = calculateHashCode
 
   protected def calculateHashCode: Int = {
-    val h0 = stringHash(getClass.getName)
+    val h0 = stringHash("org.clulab.odin.Mention")
     val h1 = mix(h0, labels.hashCode)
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
Some tests are failing...

kwalcock:
Interesting.  It passed on one version of Scala but not the other.  This version fails processors but passes reach.  The other way passed processors but failed reach.  I'll get to the bottom of it soon.

kwalcock:
More interesting.  This fails in TestTokenizers on code like
```
    println("coreTime = " + coreTime)
    println("cluTime = " + cluTime)

    // TODO: this is true when the test is run as standalone; but fails when run part of "sbt test"
    (coreTime > cluTime) should be (true)
```
The values for sbt 2.12.4 were
```
coreTime = 122
cluTime = 215
```
On sbt 2.11.11, it got
```
coreTime = 92
cluTime = 22
```
I think this has happened before and resubmitting has produced, partly by chance, values that satisfied the condition.  Perhaps the server just got busy at an untimely moment.  I'm rerunning the tests.

kwalcock:
The timing constraint was met on the second run without any code changes.

MihaiSurdeanu:
Thanks!

PR #320, opened 2019-03-19
Clean up load of x2i

kwalcock:


kwalcock:
This didn't look as good in the morning.  It's much cleaner now.

PR #323, opened 2019-04-30
Supporting  `@trigger:Label` notation for triggers

### LINKED ISSUE #321: support `@trigger:Label` notation for creating event mentions
We currently only support the `(?<trigger> ... )` notation for creating event mentions with surface patterns. This is because we only check for named intervals here: https://github.com/clulab/processors/blob/01cb3c94118645d230bcafbfc7e35b2b3e096a61/odin/src/main/scala/org/clulab/odin/impl/Extractor.scala#L49

It would be nice to support the `@trigger:Label` notation too by checking the named mentions too. See: https://github.com/clulab/processors/blob/01cb3c94118645d230bcafbfc7e35b2b3e096a61/odin/src/main/scala/org/clulab/odin/impl/TokenPattern.scala#L31-L32

BeckySharp:
Added test to verify -- previously failed (as expected), now passes.
IDK if this syntax is efficient / in keeping with odin stuff but thought it would at least get the convo started...
if merged, I think it...

closes #321

### DIFF
--- a/odin/src/main/scala/org/clulab/odin/impl/Extractor.scala
+++ b/odin/src/main/scala/org/clulab/odin/impl/Extractor.scala
@@ -45,29 +45,46 @@ class TokenExtractor(
     action(mentions, state)
   }
 
-  def mkMention(r: TokenPattern.Result, sent: Int, doc: Document): Mention =
-    r.groups.keys find (_ equalsIgnoreCase "trigger") match {
-      case Some(triggerKey) =>
+  def mkMention(r: TokenPattern.Result, sent: Int, doc: Document): Mention = {
+    val groupsTrigger = r.groups.keys find (_ equalsIgnoreCase "trigger")
+    val mentionsTrigger = r.mentions.keys find (_ equalsIgnoreCase "trigger")
+    (groupsTrigger, mentionsTrigger) match {
+      case (Some(groupTriggerKey), Some(mentionTriggerKey)) =>
+        // Can't have both notations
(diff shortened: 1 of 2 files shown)


odin/src/main/scala/org/clulab/odin/impl/Extractor.scala
```diff
@@ -45,29 +45,46 @@ class TokenExtractor(
...
-  def mkMention(r: TokenPattern.Result, sent: Int, doc: Document): Mention =
-    r.groups.keys find (_ equalsIgnoreCase "trigger") match {
-      case Some(triggerKey) =>
+  def mkMention(r: TokenPattern.Result, sent: Int, doc: Document): Mention = {
+    val groupsTrigger = r.groups.keys find (_ equalsIgnoreCase "trigger")
+    val mentionsTrigger = r.mentions.keys find (_ equalsIgnoreCase "trigger")
+    (groupsTrigger, mentionsTrigger) match {
+      case (Some(groupTriggerKey), Some(mentionTriggerKey)) =>
+        // Can't have both notations
+        throw new RuntimeException("Can't specify a trigger in as both named capture and named mention")
```
BeckySharp: this is something i don't love -- i feel like this case could be handled more cleanly?  or is it ok to throw an exception?  if people specified 2 triggers in two ways, i guess they are likely confused about how event mentions work, right?  so maybe crashing is ok? even if so, it should prob be a diff kind of exception...?

PR #328, opened 2019-05-14
Add comments to CompactWord2Vec

kwalcock:
Eidos will soon be converted to use this version.  Related code changes may be required.


BeckySharp:
thanks!!

PR #335, opened 2019-07-03
Added document attachments

MihaiSurdeanu:


MihaiSurdeanu:
@kwalcock, @bethard: is this sufficient to add the DCT to the Document class?

kwalcock:
It may need to be marked Serializable, but I'll double check when adding code for that (in DocumentSerializer and/or JSONSerializer) and then hopefully a unit test to make sure it works.  In my experience Java likes to fail silently during serialization and then you don't realize that the values are missing without a test.

MihaiSurdeanu:
Good catch. Thanks!

kwalcock:
There may need to be a way to query the keys to the document attachment map, which is protected.  JSON serialization is performed from outside the class and doesn't have direct access to the map.  Serializers may have ask for the keys, then query for each of the values, and finally serialize all the pairs.  There might eventually be other interested parties that need to know what all is in there.

bethard:
Is there a reason not to use the existing `Attachment` trait in processors?

kwalcock:
Possibly to distinguish it from a [Mention]Attachment?  If ever the two kinds are placed into the same situation, then they can be distinguished by type.  It might also be done for documentation purposes.

bethard:
One appealing thing about not having two different types is that we could use the same time of attachment for a document time as we do for a mention time.

kwalcock:
If Mihai doesn't change his mind, I *think* it would only mean doing something like this to get both:

class DCTime(val dct: DCT) extends ContextAttachment(dct.text, dct) with DocumentAttachment

MihaiSurdeanu:
I think we might have to have 2 different types of attachments, because the serializations of doc attachments vs. mention attachments might be different...

bethard:
@MihaiSurdeanu Why would the serializations be different? I mean, the individual implementations of `Attachment` would have different serializations (they already do), but that's true regardless of whether there are 2 interfaces or 1. You're imagining that a Time attached to a Document would need a different serialization than a Time attached to a mention?

@kwalcock Yeah, that might be possible, as long as there aren't conflicting method signatures. But if both `Attachment` traits are going to have serialization methods, I suspect there will be conflicts.

MihaiSurdeanu:
@bethard: yes, that's what I'm afraid of. Simply because the JSON format might look different for a doc vs. a mention.

bethard:
@MihaiSurdeanu So you think the serialization APIs might take different arguments? I wonder if it would be a good idea to get some of those serialization APIs in before (or as part of) this change? I'm having a hard time imagining what the differences might be, and having even some straw-man APIs in there might clarify your concerns.

MihaiSurdeanu:
Not necessarily different arguments, but produce different JSON. Imagine one calling the top JSON element "document-attachment" and the other "mention-attachment". Something along these lines.

bethard:
That seems like something we could avoid since we control the format. But I don't care enough about it to push for simplicity any further.

kwalcock:
My first draft is submitted above.  I'm going to try out a locally published version with the eidos DCTs and see how well it works.  Comments are welcome at any time, though.

kwalcock:
I forgot to unescape document attachments expressed as text, for instance.  I've changed a few class names.

main/src/main/scala/org/clulab/processors/Document.scala
```diff
@@ -24,12 +31,15 @@ class Document(val sentences: Array[Sentence]) extends Serializable {
   /** The original text corresponding to this document, if it was preserved by the corresponding processor */
   var text: Option[String] = None
 
+  /** Map of any arbitrary document attachments such as document creation time */
+  protected var attachments: Option[mutable.HashMap[String, DocumentAttachment]] = None
```
bethard: Maybe this is just processors style, but why use an `Option` to mean "no attachments", rather than just an empty `Map`?

main/src/main/scala/org/clulab/processors/Document.scala
```diff
@@ -78,3 +113,108 @@ object Document {
...
+class ObjectDocumentAttachmentBuilderFromText extends DocumentAttachmentBuilderFromText {
+
+  /**
+    * This matches the format in DocumentAttachment.toDocumentSerializer.
+    */
+  def mkDocumentAttachment(text: String): DocumentAttachment = {
+    val byteArray: Array[Byte] = text.split(',').map { value =>
+      val byte: Byte = java.lang.Byte.decode(value)
+      byte
+    }
```
bethard: I believe these 4 lines can be written as `val byteArray = text.split(",").map(java.lang.Byte.decode)`, and similarly when these lines of code are repeated below.

main/src/main/scala/org/clulab/processors/Document.scala
```diff
@@ -58,6 +68,31 @@ class Document(val sentences: Array[Sentence]) extends Serializable {
...
+  def addAttachment(name: String, attachment: DocumentAttachment): Unit = {
+    if (attachments.isEmpty)
+      attachments = Some(new mutable.HashMap[String, DocumentAttachment]())
+    attachments.get += name -> attachment
+  }
+
+  /** Retrieves the attachment with the given name */
+  def getAttachment(name: String): Option[DocumentAttachment] = {
+    if (attachments.isEmpty) None
+    else attachments.get.get(name)
```
bethard: I think these two lines can be rewritten as `attachments.flatMap(_.get(name))`.

main/src/main/scala/org/clulab/processors/Document.scala
kwalcock: Thanks, I'll check.  Scala was doing too many or too few conversions and this *seemed* to be the way to avoid it.

main/src/main/scala/org/clulab/processors/Document.scala
kwalcock: I will fix Mihai's code ;-)  Thanks for the careful reading.

main/src/main/scala/org/clulab/processors/Document.scala
kwalcock: I blame those mathematicians who are always distinguishing between the null set and the empty set.  I think it is to follow processors style though.

kwalcock:
Eidos is ready to use these changes, so barring objection I will merge this pull request after it is tested and then publish an update to Maven.

PR #339, opened 2019-08-23
Save and load the LstmCrfMtlParameters

kwalcock:
Notes will be added via the GitHub UI.

### DIFF
--- a/main/src/main/scala/org/clulab/sequences/LstmCrf.scala
+++ b/main/src/main/scala/org/clulab/sequences/LstmCrf.scala
@@ -345,6 +345,9 @@ object LstmCrf {
       LstmUtils.save(printWriter, rnnParameters.w2i, "w2i")
       LstmUtils.save(printWriter, rnnParameters.t2i, "t2i")
       LstmUtils.save(printWriter, rnnParameters.c2i, "c2i")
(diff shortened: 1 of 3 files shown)


MihaiSurdeanu:
That was quick...
Thanks for moving things to LstmUtils. I forgot to mention it in the email.
This looks good!

kwalcock:
Because of the similarity with LstmCrf code, some methods in LstmCrfMtl were moved around for parallel structure.  Common code was then moved to LstmUtils.

The old code still unnecessarily saves i2t even though that's not necessary.  Fixing it will invalidate model files, so that was not yet changed.

The x2i maps are now sorted upon serialization so that files match on round trips.

The parallel structure was finally abandoned in part because of `var model:LstmCrfParameters = _` and then some client-like code that needed to remember to call initialize after construction and things like that.  The constructor of LstmCrfMtl handles all the initialization now.  I didn't rearrange LstmCrf, though, in case the changes are not desired.

The test() on mtl and then on mtlFromDisk get exactly the same answers and the models "mtl" and "mtl2" are exact matches,

MihaiSurdeanu:
Thanks! Merged.

- There are no valuable models at this point, so you can go ahead and remove i2t from the model.
- You are welcome :) to rearrange LstmCrf as well!

PR #343, opened 2019-08-31
moved regexes to class level to cache compilation of patterns

MihaiSurdeanu:


kwalcock:
FWIW on a test of 1000000 iterations and only because there was already a TODO:
```scala
"""^(\.?)(\s*)\n(\s*)\n(\s*)$""".r.findFirstIn(text).isDefined // 737ms
```
```scala
val regex = """^(\.?)(\s*)\n(\s*)\n(\s*)$""".r
regex.findFirstIn(text).isDefined // 222ms
```
```scala
val pattern = """^(\.?)(\s*)\n(\s*)\n(\s*)$""".r.pattern
pattern.matcher(text).matches // 162ms
```

Since apostrophes are relatively rare, adding an
```scala
if (input.contains('\''))
```
in front of everything probably saves more time than anything.

kwalcock:
The test was failing.  The change from
```scala
      else if ("""[nN]'[tT]$""".r.findFirstIn(input.raw).isDefined) {
```
to
```scala
      else if (wont.findFirstIn(input.raw).isDefined) {
```
may not be right.  `wont` should maybe be `not`.

MihaiSurdeanu:
Thanks for speed analysis! Should we switch to caching .r.pattern then?
Also, I am not sure I understand your explanation for why the tests fail...

kwalcock:
#345 might fix the test.  I would stop there for now, though.  Maybe on a rainy day someone could optimize it locally or during some spring cleaning review all uses of regular expressions in the code base.  The very informal timings were to make sure it was headed in the right direction.

Other possibilities are to move the variables to the companion object, perform case insensitive matching, drill one step down to the pattern, drill two steps down to the matcher and the reset method, make a temporary variable raw = input.raw, incorporate the best practice into a utility class, etc.  One could get carried away.

PR #347, opened 2019-09-03
Update CHANGES.md for 7.5.4.

kwalcock:


kwalcock:
Based on the discussion at the DART meeting that just ended, I'll go ahead and merge this PR.  After that I think we're waiting on one more group to make a few changes before the DART documents are processed.

PR #349, opened 2019-09-06
We now can read the dynet model from a jar!

MihaiSurdeanu:


MihaiSurdeanu:
@kwalcock: this work now! Thanks!

kwalcock:
Nice.  It might eventually need modification for Windows which will probably want backslashes, colons, etc.  I'll break it (in) soon enough.

PR #350, opened 2019-09-09
CluProcessor is now implemented using multi-task learning and DyNet

MihaiSurdeanu:


kwalcock:
I'm testing on Windows now and also making sure the model files get closed and whatnot.

MihaiSurdeanu:
@kwalcock: so, OK to merge?

kwalcock:
I looked through all the changes, but only 100ms at each one, which was only about enough time to catch my name.  Thanks!  It's OK with me as long as it doesn't go into a pacemaker.  This being important but sort of hidden code that supports lots of other projects and part of an NLP pipeline that people need to be familiar with, it made me wonder whether there should be some regular attention paid to it by, for example, people new to the lab or new to Scala or just busybodies like me, so that more than one person has scrutinized each line of code.  That shouldn't hold back this PR though.  It's a cool use of DyNet.

MihaiSurdeanu:
Yeah, I would appreciate your constant review of this code.
Other senior people (Marco, Gus, Dane) have looked at processors code before, so most of it has been seen my multiple people. But we should make sure that this continues.

PR #351, opened 2019-09-10
Update jarred resource code

kwalcock:
Make sure it works on Windows, with %20 characters, and loose resources


kwalcock:
I verified that it also works with a zip file.  The protocol remains "jar".

PR #379, opened 2020-04-11
Add SEQ_OF_UNICODES to OpenDomainLexer.g

kwalcock:


MihaiSurdeanu:
Can you please add a comment in CHANGES as well?

kwalcock:
This was published locally and then used for reach.  The issue with the bad treatment of a UTF-32 character seemed to be solved.  I'm rerunning the reach tests, but they seem to hang on
```
21:27:37.261 [pool-6-thread-2] INFO  o.c.p.m.DeepLearningPolarityClassifier - Loading saved model SavedLSTM_WideBound_u_tag ...
```
Does something in reach need to change?

MihaiSurdeanu:
they work fine on my laptop...

kwalcock:
The problem is that git decides that the model file is a text file instead of binary and fools with the line endings so that the offsets are incorrect.  I should add the file to gitattributes so that it doesn't happen.

kwalcock:
I need to test that last one still.

kwalcock:
I put that on the wrong project.  It belongs on reach.  Fixing...

bgyori:
This is awesome, just tried this off of this branch, and the test I mentioned in https://github.com/clulab/reach/issues/662 now passes! Thanks, @kwalcock.

MihaiSurdeanu:
Thanks @kwalcock!

So, the .gitattributes should be added for all model files? Maybe I should add them to processors-models too?

kwalcock:
If the models are managed with git, then yes, to be on the safe side.  If the are only accessed as a library dependency, then it's not necessary.  The factuality model, for example, doesn't need it.  Also the x2i kinds of files really are text and get read line by line, so they don't need it.  It's just the dynet models that are small enough for git.

MihaiSurdeanu:
Got it. Thanks!

PR #381, opened 2020-04-15
add weighted embd methods

zhengzhongliang:
1, Added two functions in the CompactWord2Vec file: addWeighted and makeCompositeVectorWeighted. They allow giving different weights to the tokens when composing the embedding. 
2, The local publish version is "8.0.2-SNAPSHOT". If you modify this, the "procVer" value in eidos build.sbt should also be changed.


zhengzhongliang:
I changed back to the zipped method because iterables could not be accessed by index.

kwalcock:
I guess that alternatives are

```scala
    val weightIterator = weights.iterator

    text.foreach { word =>
      map.get(word).foreach { index => addWeighted(total, index, weightIterator.next) }
    }
```
or
```scala
    val textIterator = text.iterator
    val weightIterator = weights.iterator

    while(textIterator.hasNext && weightIterator.hasNext)
      map.get(textIterator.next).foreach { index => addWeighted(total, index, weightIterator.next) }
```
I wonder whether a timing test is worth it.

PR #385, opened 2020-04-19
Allow vars to be formatted as a yaml list...

BeckySharp:
The list elements will be combined with an `|`

This will allow for cleaner diffs when changing grammar vars/triggers etc. (hooray!)

FYI @kwalcock -- once this is merged/released I'll reformat the eidos/lum-eidos triggers to use this format, which should make our lives +100


kwalcock:
Superlative!

PR #387, opened 2020-04-23
Get Artifactory into the POM file

kwalcock:


kwalcock:
This PR add a few lines

```scala
pomIncludeRepository := { (repo: MavenRepository) =>
  repo.root.startsWith("http://artifactory.cs.arizona.edu")
}
```
 into the build.sbt file of processors-main which result in this additional output in the POM file when processors-main is published:

```xml
    <repositories>
        <repository>
            <id>Artifactory</id>
            <name>Artifactory</name>
            <url>http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release/</url>
            <layout>default</layout>
        </repository>
    </repositories>
```
This seems to be enough information for maven to figure out how to load processors-models for projects that depend on that.  It should not be necessary for such (maven) projects to worry about the transitive dependency.  sbt, however, does not seem to be as smart and was once documented as such.  I'm still testing, though, to verify.

MihaiSurdeanu:
Awesome @kwalcock !

Please ping when I can release.

kwalcock:
It indeed doesn't appear to help the situation with sbt.  With maven it at least *seemed* to: I could trick it into working.  By all accounts so far, the PR doesn't make things worse, though.  I'd go ahead and release with fingers slightly crossed.

MihaiSurdeanu:
Ok. merged, and release in process.

PR #393, opened 2020-05-22
Open the file just once when culling vectors

kwalcock:


kwalcock:
I *think* this is failing because srl5 was already failing.  This makes use of the buffering of lines that you asked about so that files don't need to be opened twice in order to process the first and remaining lines differently.  It would be nice to get this into srl5 so my hack does not propagate further.

kwalcock:
This has been resubmitted using a base branch that passes.  The code will be merged, but not with this PR.

PR #395, opened 2020-06-05
Added the Metal framework; rewrote CluProcessor with Metal

MihaiSurdeanu:


MihaiSurdeanu:
@kwalcock : can you please take a look at why this PR fails?

MihaiSurdeanu:
This test fails with "sbt test":

[info] ConstEmbeddingsGlove
[info] - should not be empty *** FAILED ***
[info]   300 was not equal to 5 (TestConstEmbeddingsGlove.scala:14)

But it passes when run individually...
@kwalcock: can you please take a look?

kwalcock:
Yes, when this new test run is complete.

kwalcock:
Unless something can be done about it, tests are run as if they are part of one big program.  TestConstEmbeddingsGlove fails because by the time it tries to make its singleton, TestLiblinearClassifier has already arranged for the singleton and that's why the dimension is 300 instead of 5.  There are lots of options for parallelism in testing, but I haven't yet found a way to tell sbt to test one file separately.  One workaround is to allow for unsetting the singleton so something else running serially can later set it again.

MihaiSurdeanu:
I see. Very nice catch! Thanks!

Given how dumb this test is, maybe the simplest option is to simply disable it?

MihaiSurdeanu:
We could add functionality for unsetting it. But then another test that may run in parallel and that expects the big file would crash...

MihaiSurdeanu:
@kwalcock : i disabled that test. Thanks!

MihaiSurdeanu:
@kwalcock : I merged your branch in here, and re-enabled that test. Let's see.

PR #396, opened 2020-06-05
Group the tests

kwalcock:
This is to avoid problems with singletons

### DIFF
--- a/main/build.sbt
+++ b/main/build.sbt
@@ -1,3 +1,6 @@
+import sbt.Tests.Group
+import sbt.Tests.SubProcess
+
 name := "processors-main"
 
 // for processors-models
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
I disabled TestConstEmbeddingsGlove in srl5. Can you please reenable it here?
If this PR passes then, can you please merge it into srl5? srl5 currently passes with this test commented out.

kwalcock:
I can't easily test this fully under Windows, so it needs to go through Jenkins first.  This code is borrowed mostly from Eidos, but InProcess was changed to SubProcess.  It *seemed* to help.  See also https://stackoverflow.com/questions/15798341/how-to-fork-the-jvm-for-each-test-in-sbt.

PR #397, opened 2020-06-05
See if this clears up the singleton tests

kwalcock:
This is temporarily headed to master just to so that it gets tested by Jenkins


kwalcock:
Nevermind.  It just took a little while for the yellow dot to show up.

PR #398, opened 2020-06-07
Buffer lines properly when culling vectors

kwalcock:
The file does not need to be opened twice


kwalcock:
There's that little lines buffering fix on a clean branch.  It was approved before, so I'm planning to merge it myself.

PR #400, opened 2020-06-19
Experiment with parallelism

kwalcock:
This is not meant to be merged.  The PR is only to facilitate discussion.


kwalcock:
There seems to be very little benefit to adding parallelism.  This is on a run of just 10 smaller files.  Parallelism is applied across the sentences of a document.  Code is included in case I misunderstood, which isn't unlikely.

Threads | Time
-- | --
1 | 0:00:01:03.526
2 | 0:00:01:01.618
4 | 0:00:00:59.918
8 | 0:00:00:58.602

MihaiSurdeanu:
Just like in Reach, where we parallelize over documents (papers).

MihaiSurdeanu:
I think this is simpler and more common. And, also, it might give us a better usage for parallelism, as the portion spend inside the DyNet synchronized block should be small(er) relative to the whole time processing a document. Hopefully this allows the threads to interleave better...

kwalcock:
Please note the introductory comment about not intending to merge.  The first version I made parallelized the processing of documents, but it crashed right away, so I decided it was probably not what you meant.  Should I look into it more carefully and probably synchronize (unparallelize) any problem areas internal to processors?

kwalcock:
Based also on threads elsewhere, I will make sure it is thread safe for documents being processed in parallel.  At worse that would mean completely synchronizing tagPartsOfSpeech, recognizeNamedEntities, and srl.  I can start there and narrow down the critical sections.

MihaiSurdeanu:
No, you don't have the synchronize anything! All the sync happens already in the Metal package, right before the calls to DyNet. You should be able to safely reuse the same CluProcessor object in parallel.

MihaiSurdeanu:
I am curious why that crash happened. It should not. Do you have a simple example that replicates it?

kwalcock:
On Windows I get `Process finished with exit code -1073740791 (0xC0000409)` just after the first document is processed.  This may indicate a stack overflow.  It could be that I need to compile with more space dedicated to the stack, especially if results differ on some other OS, but it usually means there's some underlying problem.  I have not yet verified on a server or looked into it closely.  I'm using `ParallelProcessorExample.scala` in branch `kwalcock-parallel` with arguments `../corpora/Doc52/txt ../corpora/Doc52/out txt 1` and then >1.

MihaiSurdeanu:
Can you please create a small unit test for this?

kwalcock:
Will do, forthcoming.

kwalcock:
There is a unit test version in the kwalcock-parallel branch file TestParallel.scala that  can be run with `sbt` as `corenlp/testOnly org.clulab.processors.TestParallel`.  The result on Amy is that java/sbt crashes:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fc265e1f0b9, pid=24364, tid=0x00007fc2779f9700
#
# JRE version: Java(TM) SE Runtime Environment (8.0_131-b11) (build 1.8.0_131-b11)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.131-b11 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [dynet9047127878655664153.dll+0x12b0b9]  dynet::SimpleExecutionEngine::incremental_forward(unsigned int)+0x599
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /net/kate/storage/work/kalcock/Projects/processors/hs_err_pid24364.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
```

MihaiSurdeanu:
Thanks! I will take a look. This does sound like a synchronization issue.

MihaiSurdeanu:
Hi @kwalcock: this bug was fixed. It was a synchronization issues, and it was my fault. The test completes successfully now and produces the same output as the sequential run. This should be merged in master asap, as this is a serious bug.

MihaiSurdeanu:
When you measure runtimes, you should exclude:
- The time to load the word embeddings in ConstEmbeddingsGlove.apply here:
https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/dynet/ConstEmbeddingsGlove.scala#L71

- The times to load the Metal models here:
https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/dynet/Metal.scala#L492

Thank you!

MihaiSurdeanu:
If it's Ok with you, can you please merge this in master before the runtime experiments? That bug is big. And thanks for finding it!

kwalcock:
Yes

kwalcock:
Merging, I guess.  There will be a slight adjustment in the timing next.

MihaiSurdeanu:
Yes, thanks!

MihaiSurdeanu:
Thanks!

I am curious about timing measurements. It may be that threading won't help... But at least we are thread safe now.

kwalcock:
The safely part is much appreciated, even if it isn't helping much with the time.  I just had to change the example "priming" sentence to one that included some predicates in order to get the last Metal model to load before the timer began.

Threads | Time
-- | --
1 | 0:00:00:54.215
2 | 0:00:00:55.925
4 | 0:00:00:53.910
8 | 0:00:00:53.763

MihaiSurdeanu:
Yeah, it seems that DyNet indeed does not parallelize well due to the static computation graph... Oh well...

PR #420, opened 2020-09-09
Propagate case setting to IntHashTrie

kwalcock:


kwalcock:
I *hope* this is the extent of the problem.  Thank you, @gcgbarbosa, for finding it.  If the LexiconNERShell should make it to master, then this branch can be merged with lexner and the result merged to master.  If not, then I should make a separate PR on master with just this change.

gcgbarbosa (approved):

gcgbarbosa:
@kwalcock thanks a lot!

MihaiSurdeanu (approved):

MihaiSurdeanu:
Thanks @kwalcock !

I think we need to release processors, and update reach, because this bug impacts reach. When you have a chance, can you please do it?

kwalcock:
Processors is republished.  Reach is to follow.  Sorry about the bug.  It's probably not the only one.

MihaiSurdeanu:
Thank you!

PR #421, opened 2020-09-09
Get case bug fix into master

kwalcock:


kwalcock:
Thanks for updating the change file, @MihaiSurdeanu.

PR #423, opened 2020-09-17
Update for new version of fatdynet

kwalcock:


MihaiSurdeanu:
Thanks!

I changed the clone() to "this" in Greedy and ViterbiForwardLayer in the "parallel" branch, so your two changes might cause a merge conflict. But it will be easy to solve.

kwalcock:
It looks like FatDynet is at maven central, so when I push merge, the tests should start passing.

PR #427, opened 2020-10-03
Use specific problematic text

kwalcock:


kwalcock:
The grounder in eidos annotates the entire document as does this test program.  However, the grounder then takes each individual sentence as a document text and annotates that.  This text comes from one of the other documents that is crashing and that I happened to find first.  The trailing \n\n means that in the larger document text might be getting an extra end of paragraph period there which might explain why this test program wasn't having a problem.  The test now breaks for me.  I'll look for the sentence in document 3 as well.

MihaiSurdeanu:
Ok, I can confirm the fail now. Thanks!

PR #428, opened 2020-10-03
Finally fixed incorrect dep graph sizes

MihaiSurdeanu:


MihaiSurdeanu:
@kwalcock: I believe this finally fixed that Eidos crash. Thanks for the examples!

kwalcock:
This did seem to fix the problem.  The file 3_ALPAN_Paper_10.txt could be removed now.  Can this version be sent to maven?  It might be the version just before dynet goes parallel and it is a good fallback position if parallel goes wrong.  If it doesn't get published, people (I) would need to use the locally published version to get the fix.

MihaiSurdeanu:
Good idea. Can you please release this?

kwalcock:
It was just released.

MihaiSurdeanu:
Nice, thanks!

Issue #424, opened 2020-09-18
java.lang.NumberFormatException: For input string: "<STOP>"


kwalcock:
Running on branch parallel and in parallel with CluProcessor, this exception occurs during annotation.  I will attach a file if/when it can be reproduced.

```
Processing 9_FAOSS Livestock Crisis Update 2015.txt...
Processing 32_sudantribune.txt...
Processing 33_Achievement of household food security in South Sudan - Sudan Tribune_ Plural news and views on Sudan.txt...
Processing 31_South_Sudan_2018_Humanitarian_Needs_Overview.txt...
Processing 11_FAO_a-i5591e.txt...
Processing 12_SSD21-February-2017.txt...
Processing 26_South_Sudan_2018_Humanitarian_Needs_Overview.txt...
Processing 27_South Sudan_ Floods - Sep 2017 _ ReliefWeb.txt...
Processing 4_FAO.txt...
Processing 50_wfp280481.txt...
[error] (run-main-0) java.lang.NumberFormatException: For input string: "<STOP>"
[error] java.lang.NumberFormatException: For input string: "<STOP>"
[error]         at java.lang.NumberFormatException.forInputString(Unknown Source)
[error]         at java.lang.Integer.parseInt(Unknown Source)
[error]         at java.lang.Integer.parseInt(Unknown Source)
[error]         at scala.collection.immutable.StringLike.toInt(StringLike.scala:301)
[error]         at scala.collection.immutable.StringLike.toInt$(StringLike.scala:301)
[error]         at scala.collection.immutable.StringOps.toInt(StringOps.scala:29)
[error]         at org.clulab.processors.clu.CluProcessor.$anonfun$parseSentence$3(CluProcessor.scala:199)
[error]         at scala.runtime.java8.JFunction1$mcVI$sp.apply(JFunction1$mcVI$sp.java:12)
[error]         at scala.collection.TraversableLike$WithFilter.$anonfun$foreach$1(TraversableLike.scala:789)
[error]         at scala.collection.immutable.Range.foreach(Range.scala:156)
[error]         at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:788)
[error]         at org.clulab.processors.clu.CluProcessor.$anonfun$parseSentence$1(CluProcessor.scala:198)
[error]         at org.clulab.processors.clu.CluProcessor.$anonfun$parseSentence$1$adapted(CluProcessor.scala:193)
[...]

PR #432, opened 2020-10-27
Prepare to publish 8.2.3

kwalcock:
Updated CHANGES
Fixed bug in issue #424

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 # Changes
-+ **8.2.2** - Do not allow loops in enhanced semantic roles.
++ **8.2.3** - Do not allow loops in enhanced semantic roles.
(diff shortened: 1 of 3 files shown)


kwalcock:
Eidos used the changes committed to master (since 8.2.2) but not yet published plus this other bug fix which was only available in the parallel branch for the latest reading.  Several changes related to the reading are being committed in eidos and headed for the master branch there.  That doesn't work if eidos is dependent on a locally published processors.  I'd like to commit and then publish these changes as 8.2.3 for the sake of eidos and anything else that might need it.  8.2.4 or 8.3.0 would then get the changes from the parallel branch.

MihaiSurdeanu:
Please go ahead!

PR #434, opened 2020-11-18
Make sure ComputationGraph construction is synchronized

kwalcock:
This is largely for documentation purposes.  Maybe it will prevent some hard-to-debug situation in the future.


MihaiSurdeanu:
Interesting. I wonder if this makes a difference...

kwalcock:
I don't think it does, but the next PR will adjust destruction a little and it seems to help.

MihaiSurdeanu:
Oooo. Exciting 👍

kwalcock:
I don't want to merge this (yet if ever), but I think some kind of defensive programming like this could be useful.  The computation graph is only addressed from this changed code and not outside it.  Synchronization is also handled here and it may be on the global object or later on a thread or even cg itself.  An experiment with this showed that ComputationGraph.renew is not necessary if clear is used.  Clear will reset the graph so that it can be used again without Java doing garbage collection on the abandoned cg.  Clients of Metal don't need to be renewing their cg.

kwalcock:
Although the change helped convince me that processors is handling the cg correctly, I'm still getting a crash on an AWS machine, but so far nowhere else.  It is repeatable but unfortunately takes 20 or 30 minutes.  Then the program quite summarily crashes.  I'm still trying to find out why.

kwalcock:
These changes were copied in from a different branch without complete integration.  Please ignore failing tests for now.

MihaiSurdeanu:
I think this should be merged. It is very clean code that forces us to behave well...

PR #435, opened 2020-11-29
Make sure CG construction is synchronized

kwalcock:
Use block for CG operations
Check CG version before and after

### DIFF
--- a/main/src/main/scala/org/clulab/dynet/ConstEmbeddingsGlove.scala
+++ b/main/src/main/scala/org/clulab/dynet/ConstEmbeddingsGlove.scala
@@ -69,13 +69,10 @@ object ConstEmbeddingsGlove {
   private var SINGLETON: Option[ConstEmbeddingsGlove] = None
 
   def apply(matrixResourceName: String, isResource: Boolean): ConstEmbeddingsGlove = {
-
-    DyNetSync.synchronized {
-      // these objects are read-only and they use a lot of RAM, so let's reuse them if they exist
-      if(SINGLETON.isEmpty) {
-        SINGLETON = Some(new ConstEmbeddingsGlove(matrixResourceName, isResource))
-      }
-      SINGLETON.get
-    }
+    // These objects are read-only and they use a lot of RAM, so let's reuse them if they exist.
+    // No ComputationGraph is touched during this process, so synchronization is not required.
+    if (SINGLETON.isEmpty)
+      SINGLETON = Some(new ConstEmbeddingsGlove(matrixResourceName, isResource))
+    SINGLETON.get
   }
 }
--- a/main/src/main/scala/org/clulab/dynet/DyNetSync.scala
+++ b/main/src/main/scala/org/clulab/dynet/DyNetSync.scala
@@ -1,6 +1,59 @@
 package org.clulab.dynet
 
+import edu.cmu.dynet.ComputationGraph
+
 /** Use this object for synchronized statements in DyNet */
 object DyNetSync {
+  protected var expectedVersion = 0
 
+  def withComputationGraph[T](f: => T): T = {
+    // In parallel version, synchronize on Thread.currentThread.
+    this.synchronized {
+      try {
+        require(ComputationGraph.version == expectedVersion)
+        val result = f
+        require(ComputationGraph.version == expectedVersion)
+        result
+      }
+      finally {
+        // Make sure the nodes are freed immediately.  This prevents live object
+        // from being trashed and may help prevent memory fragmentation.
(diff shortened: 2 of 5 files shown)


MihaiSurdeanu (approved):

kwalcock:
I changed some code in the Utils class near this comment because of its proximity to the crash site:

// Some embeddings may be empty in some weird Unicode encodings

Can you explain briefly what was happening here.  How can outs be empty?  It might be a clue.  I don't think the execution ever takes this path, but the crash dump does mention an ExpressionVector and "(0 bytes)" just before the crash each time:

`Java frames: (J=compiled Java code, j=interpreted, Vv=VM code)
J 21119  edu.cmu.dynet.internal.dynet_swigJNI.concatenate__SWIG_1(JLedu/cmu/dynet/internal/ExpressionVector;)J (0 bytes) @ 0x00007f6257ec6455 [0x00007f6257ec6400+0x55]`

```scala
  def mkCharacterEmbedding(word: String,
                           c2i: Map[Char, Int],
                           charLookupParameters: LookupParameter,
                           charFwRnnBuilder: RnnBuilder,
                           charBwRnnBuilder: RnnBuilder): Expression = {

    def safelyTransduce(charEmbeddings: Seq[Expression], rnnBuilder: RnnBuilder): Expression = {
      val outs = transduce(charEmbeddings, rnnBuilder)
      val out =
        if (outs.nonEmpty) outs.last
        // Some embeddings may be empty in some weird Unicode encodings.
        else {
          logger.warn(s"A strange character was encountered in word '$word'.")
          val altCharEmbeddings = Array(lookup(charLookupParameters, UNK_EMBEDDING))
          val result = transduce(altCharEmbeddings, rnnBuilder).last
          result
        }

      out
    }

    //println(s"make embedding for word [$word]")
    val charEmbeddings = word.map { c: Char =>
      lookup(charLookupParameters, c2i.getOrElse(c, UNK_EMBEDDING))
    }
    val fwOut = safelyTransduce(charEmbeddings, charFwRnnBuilder)
    val bwOut = safelyTransduce(charEmbeddings.reverse, charBwRnnBuilder)

    concatenate(fwOut, bwOut)
  }
```

kwalcock:
In the code above, the former .`last` then `.first` in
```scala
      if (bwOuts.nonEmpty) bwOuts.last
      else transduce(Array(lookup(charLookupParameters, 0)), charBwRnnBuilder).head // 0 = UNK
```
was changed to last/last.

MihaiSurdeanu:
I suspect this line in Utils.mkCharacterEmbedding is problematic:

https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/dynet/Utils.scala#L365

Somehow, the `c2i` array gets out of sync with `charLookupParameters`?

MihaiSurdeanu:
`c2i` and `charLookupParameters` are initialized here:
https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/dynet/EmbeddingLayer.scala#L401-L410

BUT, if you just run processors, they are loaded from a model file here:
https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/dynet/EmbeddingLayer.scala#L255
and here:
https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/dynet/EmbeddingLayer.scala#L255
(charLookupParameters is populated when the parameters are actually populated)

I wonder if we can add a check for after everything is loaded, to make sure this situation does not happen?

kwalcock:
Yesterday I checked on the dimension of 109 in the x2i and rnn files and it all looked correct.  Did you come across a particular character or a particular file that necessitated the code, because I can't imagine including it otherwise?  Could it have been one of the malformed utf-32 characters?  I could just swamp it will all unicode characters and try to get it to break.

MihaiSurdeanu:
I never ran into this situation before...
I guess the stress test where you swamp it with all unicode chars won't hurt...

I wonder if we can get more info on that empty ExpressionVector? Can we link it to an actual var name?

kwalcock:
This PR is for the parallel processing version of FatDynet, which is going to need to be revised now that the memory problems have been addressed.

MihaiSurdeanu:
Thanks @kwalcock !
There is a merge conflict here.

kwalcock:
"This is going to need to be revised" meaning this is an ancient PR just serving as a placeholder of something yet to come.  I was reviewing PRs for all projects and should have let it be.  FatDynet is going to get another update before the parallel functionality is again available.  I filed a memory management issue with CMU yesterday and would like to see what they say.

kwalcock:
This is soon to be superseded, so I'm removing it.

PR #439, opened 2021-01-05
 Bug fix: we now guarantee that the Stanford dependency graph produce…

MihaiSurdeanu:
…d by CoreNLP has the same number of nodes as the words in the sentence.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # Changes
++ **8.2.4** - Bug fix: we now guarantee that the Stanford dependency graph produced by CoreNLP has the same number of nodes as the words in the sentence.
 + **8.2.3** - Do not allow loops in enhanced semantic roles.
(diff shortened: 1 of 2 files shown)


kwalcock:
@MihaiSurdeanu, reach is about to be republished to use the updates to bioresources.  It would also like to use a version of processors with this fix.  That means processors would need to be updated first.  Is it OK to release version 8.2.4 from this commit?

MihaiSurdeanu:
Please do!
Can you please check that CHANGES is also updated?

kwalcock:
It looks like you already had done that for processors.  I did for reach.  Ben did for bioresources.

kwalcock:
This PR was never actually tested.  River was restarted recently and it seems that it has a problem with the first test(s) after the restart.  I am trying to force it to test.  Locally, it is TestDocumentSerializer that is failing.  The recent change with the preferredSize causes serialization problems.  That value is not preserved in the process, perhaps because there are empty elements padding to the end of a container that are either not written or read.  Here is a screenshot:

image

PR #441, opened 2021-01-08
Set preferredSize when DirectedGraph is deserialized

kwalcock:


MihaiSurdeanu:
Thanks @kwalcock !

kwalcock:
This was again not properly tested by Jenkins.  I am removing the changeset check in the jenkinsfile.  It is apparently unreliable.  Neither eidos nor reach has such a thing.  An extra test run or two will not hurt much, I figure.

kwalcock:
I'm starting this over to make extra sure that the merged version is also being tested properly.

PR #443, opened 2021-01-18
Add reusable menu components

kwalcock:
There are at least four copies of the shell code in circulation: processors, reach, eidos, eidosdottir.  It was time to reuse some code.  This is being copied over from eidos and then eidos and reach will use it.  There are already PRs prepared for them.


kwalcock:
Some downstream projects will soon be updated and would like to reuse this code.  @MihaiSurdeanu 's name is on it, so he should be made aware of it.

MihaiSurdeanu (approved):
Thanks!

PR #445, opened 2021-02-07
Integrate with Fatdynet 0.3.0 for better memory management

kwalcock:
This is still being tested on multiple machines and OSes, but here's a preview.  This will provoke Jenkins to test it on river.


MihaiSurdeanu:
Thanks @kwalcock !

Can you please remind me what blocks of code like `Synchronizer.withComputationGraph("<some name>")` do?

Also, why do we need the `safelyTransduceLast1` and `safelyTransduceLast2` methods?

kwalcock:
I'm not sure I like the name withComputationGraph, but it emphasizes that manipulations of the computation graph should be limited to the block of code that follows.  withComputationGraph can track the ComputationGraph version and complain if it sees something strange.  It renew()s the CG upon completion of the block, even if exceptions are thrown.  In verbose mode the Synchronizer outputs useful information.  There's a withoutComputationGraph that might be used for DyNet things needing synchronization that don't involve the CG.  The Synchronizer is the centralized manager for threading issues like DyNetSync, except that it has been moved to FatDynet.

kwalcock:
`safelyTransduceLast1 ` and `safelyTransduceLast2` are alternative implementations of the same thing.  Only one is needed and it looks like I chose 1.  At one time both were run and the results compared.  I left both for documentation, at least temporarily.

Issue #446, opened 2021-02-07
We do not need to skip the tests marked NeedsExternalBinary


kwalcock:
Since we are testing on river, we can just install the external binary programs and then test everything.

kwalcock:
Fixed in #449.

PR #449, opened 2021-02-09
Configure NeedsExternalBinary tests dynamically

kwalcock:


kwalcock:
See also #446.  Since we control the Jenkins computer river, we can install the external binaries and run the tests, which was done here.

Scalatest issues reports with counts for tests succeeded, failed, canceled, ignored, and pending.  We have largely neglected the canceled category, which can be put to use here.  Rather than arranging for the tests requiring external programs to be ignored, we can dynamically cancel them if they are found to be unavailable at runtime.  In this way the tests can be left on and if they happen to run on a machine without the binaries, the test suite won't fail.  For instance, if a user tests locally, they won't see failed tests.

kwalcock:
But @MihaiSurdeanu could weigh in on this one.

MihaiSurdeanu:
I think this is great!

PR #450, opened 2021-02-10
Use Menu in shell

kwalcock:
Don't use the PrintWriter
Manage printWriter locally
Rearrange Shell

Rename method


kwalcock:
With #443 I did not notice that processors already has three or four other shells.  They have been updated to use the library classes added back then.  The semantics of the printWriter have changed a little and I hope that nothing depends on it.  It is now local to the CluShell which is the only one that was using it among these.

MihaiSurdeanu (approved):
Looks good! Please merge.

PR #453, opened 2021-02-13
bug fix: handle cycles robustly in dependency trees produced by CoreNLPProcessor

MihaiSurdeanu:


MihaiSurdeanu:
@kwalcock: can you please take a look at why this doesn't compile? Hopefully, this solves the headlessDep issue in reach.

kwalcock:
I'm not sure about not compiling, but this test isn't happy.

```
[info] DependencyUtils
[info] - should handle cycles in the dependencyGraph correctly
[info] - should handle roots with incoming dependencies
[info] - should throw a DependencyUtilsException if the dependency parse is malformed *** FAILED ***
[info]   Expected exception org.clulab.utils.DependencyUtilsException to be thrown, but no exception was thrown (TestDependencyUtils.scala:109)
```

MihaiSurdeanu:
I'll fix it soon.

PR #454, opened 2021-02-15
Revise root finding

kwalcock:
Set.toList has been changed to List.distinct, which may affect ordering of the result, mostly by ensuring it is consistent.  Otherwise output should be unchanged.


MihaiSurdeanu (approved):
Thanks!

PR #456, opened 2021-02-27
LexiconNER case sensitivity

kwalcock:
Allow per KB case sensitivity, automatic matching in overrideKBs, and default setting

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # Changes
++ **8.2.7** - Allow LexiconNER collection to use KBs with differing case sensitivity
 + **8.2.6** - Bug fix: handle cycles in dependency trees produced by CoreNLPProcessor.
 + **8.2.5** - Added CluCoreProcessor, which uses all our tools + Stanford's numeric entity recognizer. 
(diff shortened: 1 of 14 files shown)


MihaiSurdeanu:
There are a lot of changes here that I don't understand... This is now diverging from my original implementation in significant ways. I think we'll have to meet to walk me through all these. Let's organize a meeting on Monday.

kwalcock:
Sure.  I finally figured out how to get GitHub to show split diffs and that might help put the changes in context.  Your original implementation in SeparatedLexiconNER and BooleanHashTrie is largely unchanged.  The other two, the Combined and Compact, aren't at all new, but they relied heavily on the case sensitivity being the same for everything, so they needed to be changed significantly when that was no longer true.  Basically, they doubled up their internal structures and use one for each case sensitivity value.

Jenkins is having some kind of network problem now, but the passing tests should show up when it gets figured out.  Quite a few were added for this code.  I also ran reach on a bunch of files using all three of these NERs and they got the same answers.

kwalcock:
Scaladoc documentation has been added.  Just let me know if it needs another draft.

MihaiSurdeanu (approved):
Looks great @kwalcock ! Thanks!
Ok to merge.

PR #457, opened 2021-03-17
Use the word embedding map pool

kwalcock:
Provide InputStreamer to read resources
Update build configuration
Supply two implementations of WordEmbeddingMap
Measure and improve performance
Allow glove files to skip header
Make use of updated fatdynet


kwalcock:
50 commits worth of experiments and design drafts are squashed into this.  The messy version is in branch kwalcock/constembed in case some cherry picking needs to be done, for instance if tests fail.  I am still sorting out impact on other projects and getting the compressed version published.

MihaiSurdeanu:
Yay!
- But one build fails.
- Also, the CHANGES needs to be way more verbose for 8.3.0.

kwalcock:
There seem to be memory issues.  All these static objects are just a sneaky way of creating state and they are catching up with us.  The testing environment is extra complicated because of single or multiple JVMs involved, parallel testing at multiple levels, confusing sbt documentation, etc.

kwalcock:
```
[error] java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: GC overhead limit exceeded
```

kwalcock:
Please ignore some of the test chatter.  The small commits are partially intended to give Jenkins plenty of opportunity to run into problems.  These changes are all being tested on multiple computers because they involve time-sensitive interaction between the garbage collector and dynet.  It looks like a change I made a month ago regarding synchronization was based on an assumption that wasn't quite true.  Initialization is being invoked under more circumstances than I had realized.  It seems stable now.  The problems were not a result of very recent updates.

kwalcock:
Is it necessary to replace any of these deprecated methods with versions that accept a WordEmbeddingMap.  The last one from Oden is of particular concern.

```
[warn] D:\Users\kwa\Documents\MyData\Projects\processors-project\processors\main\src\main\scala\org\clulab\dynet\Utils.scala:865:24: class SanitizedWordEmbeddingMap in package embeddings is deprecated: ExplicitWordEmbeddingMap should replace the functionality in this class
[warn]   def mkWordVocab(w2v: SanitizedWordEmbeddingMap): Map[String, Int] = {
[warn]                        ^
[warn] D:\Users\kwa\Documents\MyData\Projects\processors-project\processors\main\src\main\scala\org\clulab\dynet\Utils.scala:875:33: class SanitizedWordEmbeddingMap in package embeddings is deprecated: ExplicitWordEmbeddingMap should replace the functionality in this class
[warn]   def initializeEmbeddings(w2v: SanitizedWordEmbeddingMap, w2i: Map[String, Int], lookupParameters: LookupParameter): Unit = {
[warn]                                 ^
[warn] two warnings found
[info] Done compiling.
[info] Compiling 27 Scala sources to D:\Users\kwa\Documents\MyData\Projects\processors-project\processors\odin\target\scala-2.11\classes ...
[info] Compiling 24 Scala sources and 1 Java source to D:\Users\kwa\Documents\MyData\Projects\processors-project\processors\corenlp\target\scala-2.11\classes ...
[info] Done compiling.
[warn] D:\Users\kwa\Documents\MyData\Projects\processors-project\processors\odin\src\main\scala\org\clulab\odin\impl\OdinResource.scala:12:51: class SanitizedWordEmbeddingMap in package embeddings is deprecated: ExplicitWordEmbeddingMap should replace the functionality in this class
[warn] class EmbeddingsResource(is: InputStream) extends SanitizedWordEmbeddingMap(is, None, false) with OdinResource {
```

kwalcock:
One of the tests that periodically fails is one with a timing.  I'm not sure whether cluTime is getting slower or we just get unlucky with garbage collection delays or some parallel processes.  Tests of only a couple of ms probably aren't very reliable.  Usually if I retest it goes away.

```
coreTime = 2
cluTime = 4
[info] - should have similar outputs *** FAILED ***
[info]   false was not true (TestTokenizers.scala:80)
```

MihaiSurdeanu:
I think it's fine to relax that time constraint to a safe value.

PR #460, opened 2021-03-23
Share the glove files

kwalcock:


kwalcock:
I thought this was already tested, but apparently not.  We've been claiming to still support Scala 2.11 for a long time, but haven't been testing it regularly.  It turns out that it doesn't work.  It's in a tricky place where Scala has changed.

kwalcock:
I take that back.  This is a new file brought over from Eidos.

PR #465, opened 2021-03-30
Fix reach tests

kwalcock:
This is to get the unit tests running.  Documentation will follow.


kwalcock:
I'm trying to figure out the red X above.  It appears to be a dynet thing again.  However, the bug fix here is pretty important and it seems worth publishing that first so that people have an alternative and don't use 8.3.0.

PR #467, opened 2021-04-10
Add unknownKey to interface

kwalcock:


kwalcock:
The compact representation actually already includes the entire vector as if it were a lookupParameter, but the indexes might be different than those in w2i.  It has its own w2i map as well.  They're practically the same thing.  If I just removed "protected", one could get at it.

MihaiSurdeanu (approved):

PR #469, opened 2021-04-13
Fixed the crash in the word embeddings in dynet

MihaiSurdeanu:


MihaiSurdeanu:
I am no longer storing the WordEmbeddingMap in the pool. Just use it to create the LookupParameter, and then discard it. This seems to save considerable memory. Tests pass on my machine.

kwalcock:
I understand the need for the change.  Processors is a library and there is no guarantee that anyone else is going to need the embeddings.  If they do, though, we're not saving space or time.  I wonder if there can be something next to the getOrElseCreate like getOrElseIgnore or get():Option that might be used just in case something is aware of the sharing potential.  That thing would need to be sure to call the getOrElseCreate before processors calls get(), though.  That's a little troublesome.

That's a good call on the synchronize.  It's probably missing on a large number of singletons.  This particular singleton probably isn't a problem, but sometimes it would be nice to get rid of them, which is an alternative to the above design.  Some method could be called to remove the singleton and then if nothing was indeed using it, it would be garbage collected.  When testing dynet there were times when I needed to clobber something like this in order to make sure that all memory was really being freed.  I wonder whether there is a singleton trait that could incorporate the unsetting and synchronization.  object ConstEmbeddingsGlove might extend the Singleton[ConstEmbeddingGlove] trait.  I will look it up and probably experiment elsewhere.

MihaiSurdeanu:
I agree with all these suggestions. Do you mind adjusting the WordEmbeddingMapPool to reflect the first paragraph?
But, all these are temporary solutions... I am very frustrated that we need to duplicate the glove vectors... I am thinking of alternatives, but I don't have a good solution. Yet.

kwalcock:
Just now this was working with the streaming version of Eidos.  I think it should be merged and published.

MihaiSurdeanu:
Yay! Thanks!

PR #470, opened 2021-04-15
Allow checking for reusable WordEmbeddingMap

kwalcock:


kwalcock:
A different procedure would to go ahead and call the getOrElseCreate in case processors is the first one to need the word embeddings but then to eventually clear it from the pool.  If nobody else has one, it will go away.  In this way there might be less of a race condition.  Maybe next time

Most other singletons are lazy vals and they are synchronized by default.  This one is a special optional singleton, but still a pattern eventually worthy of a trait.

MihaiSurdeanu:
Thanks!

PR #473, opened 2021-04-30
Add unit test

kwalcock:


MihaiSurdeanu:
Thank you!

kwalcock:
It wasn't until this morning that I confirmed that the test also fails on Linux, but at least it is consistent.  The one clue I found was that in `ConstEmbeddingsGlove.mkConstLookupParams` the line `val w2i = words.zipWithIndex.toMap` has an unnecessary `.toMap`.  Later in `for ((word, index) <- w2i)` the values will be taken from `w2i` out of order based on hash codes and bins and things.  Often it means that the order changes between runs.  Order could matter because if nodes are summed in different orders, the rounding/truncation can result in different answers.  The order seems to stay consistent in this case.  Nevertheless, when the `.toMap` is removed, the test passes.  I looked through the C code to find out what is happening, but I don't see that the order should matter at all with a lookup parameter.  IIRC correctly, the result was consistent, but different than either of the other conflicting results.  Something very strange is going on.

kwalcock:
Oops!  The/a problem is indeed here.  In this particular sentence, some words appear multiple times.  When they are placed in the map, only one of their positions can be stored.  The length of words is 17 while the size of w2i is only 14.  When the words and one of their indexes are used to initialize the vector, some of the positions in the vector remain uncovered and the memory there is left to chance.  The `.toMap` should be removed.  With that change the problem test and all others pass locally.  I have some other optional changes made while trying to figure things out.  They'll be in another PR.  I'll do the timings soon.

MihaiSurdeanu:
Ha! Beautiful catch!!! I am super dumb for not seeing this. Any CS freshman should've caught it. 
(I copied & pasted this code from code that builds a vocabulary, and couldn't be bothered to think about the change in semantics...)

MihaiSurdeanu:
Now that you fixed this, can you please do a runtime analysis? I have some ideas on how to make this more efficient, but I'm not sure it's worth it.

kwalcock:
These are times (minutes) to completion of the test program.

  | Branch | Branch
-- | -- | --
threads | master | glovecrash3
2 | 7:06 | 7:36
1 | 8:59 | 9:05

The new branch takes a little longer than the old one.  In ConstEmbeddingsGlove I am reusing the temporary FloatVector in mkConstLookupParams so that there is only one per method call.  It can be changed to a global, one per object, but then it would need to be synchronized.  This does not take into account things like reduced memory usage, long-term garbage collection. etc.

MihaiSurdeanu:
Thanks! While I work on the improvement, I think this should be merged in master.

kwalcock:
I still have to address the FIX ME parts.

kwalcock:
This test doesn't happen to include construction time for the processor.

threads | master | glovecrash3 | master | glovecrash4
--- | --- | --- | --- | ---
2 | 7:06 | 7:36 | 7:19 | 7:11
1 | 8:59 | 9:05 | 8:51 | 8:39

MihaiSurdeanu:
Nice! So, faster than master. Why are there two "master" columns?

MihaiSurdeanu:
It seems glovecrash4 should be in master? Then, finally, you can get to the merging of the other CLAB and parallelism changes...

The more exciting thing to me is that we can now optimize the WordEmbeddingMap classes as we wish, as they are disconnected from DyNet.

kwalcock:
It was just to make sure the computer is approximately as fast today as it was yesterday and to get an idea of the variance.

kwalcock:
Here are some times for the glovecrash4 branch on the same batch of documents that were run above, but this time with varying processor:

Threads | FastNLPProcessorWithSemanticRoles | FastNLPProcessor | CluProcessor
--- | --- | --- | ---
2 | 7:10 | 2:16 | 9:41
1 | 8:58 | 3:23 | 9:36

MihaiSurdeanu:
Thanks! 
On the 1 thread setting, CluProcessor is disappointingly slow. I think I know why, but, just to confirm it, can you please run 2 more versions of CluProcessor, each without one of these methods:

https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/processors/clu/CluProcessor.scala#L97

https://github.com/clulab/processors/blob/master/main/src/main/scala/org/clulab/processors/clu/CluProcessor.scala#L95

I suspect the srl() method is causing the slowdown, because it is O(n^2).

kwalcock:
The SRL takes almost half the time.

Threads | CluProcessor base | without lemmatize | without srl
--- | --- | --- | ---
2 | 9:38 | 10:01 | 5:30
1 | 9:34 | 9:50 | 5:27

MihaiSurdeanu:
Thanks! 
CluProcessor without SRL has essentially the same functionality as FastNLP. I wonder why it's slower than that. Can you please analyze the individual calls in CluProcessor.annotate() to see where the time is spent?
No rush, given the hackathon.

kwalcock:
Calls | CluProcessor | FastNLPProcessorWithSemanticRoles | FastNLPProcessor
--- | --- | --- | ---
chunking | 0:00:00:00.000 | 0:00:01:44.861 | 0:00:01:52.302
clear | 0:00:00:00.000 | 0:00:00:00.000 | 0:00:00:00.000
discourse | 0:00:00:00.000 | 0:00:00:00.000 | 0:00:00:00.000
lemmatize | 0:00:00:00.180 | 0:00:00:00.153 | 0:00:00:00.161
mkConstEmbeddings | 0:00:00:00.619 |  | 
parse | 0:00:04:39.501 | 0:00:00:27.650 | 0:00:00:27.524
recognizedNamedEntities | 0:00:00:40.219 | 0:00:01:15.281 | 0:00:01:21.083
relationExtraction |  | 0:00:00:00.001 | 0:00:00:00.001
removeAttachment | 0:00:00:00.000 |  | 
resolveCoreference | 0:00:00:00.000 | 0:00:00:00.000 | 0:00:00:00.000
srl | 0:00:04:29.960 | 0:00:05:32.942 | 0:00:00:00.000
tagPartsOfSpeech | 0:00:00:41.661 | 0:00:00:02.194 | 0:00:00:02.270
total (minutes) | 10.54 | 9.05 | 3.72

kwalcock:
It's apparently somewhere else:

Calls | CluProcessor
--- | ---
chunking | 0:00:00:00.000
clear | 0:00:00:00.000
discourse | 0:00:00:00.000
**generateUniversalEnhancedDependencies** | **0:00:00:00.183**
lemmatize | 0:00:00:00.219
mkConstEmbeddings | 0:00:00:00.833
parse | 0:00:04:07.801
recognizedNamedEntities | 0:00:00:38.760
removeAttachmen | 0:00:00:00.000
resolveCoreference | 0:00:00:00.000
srl | 0:00:04:12.360
tagPartsOfSpeech | 0:00:00:41.086

MihaiSurdeanu:
Thanks!

Is there a cmd line I can use to repeat these tests?

kwalcock:
I pushed the branch kwalcock/glovecrash4. `testOnly org.clulab.processors.TestParallel` produces the output.  The last few commits show how to use it.  The timer works reasonably for measurement, but not so good for discovery.  If it isn't soon obvious what the slow part is, then it might be necessary to break out FlightRecorder.

MihaiSurdeanu:
Thanks! So, the numbers I should be looking at are under "Serial processing complete"?

```
Serial processing complete.
Time	chunking	0:00:00:00.000	171959
Time	clear	0:00:00:00.000	52518
Time	discourse	0:00:00:00.000	48232
Time	generateUniversalEnhancedDependencies	0:00:00:00.140	140018056
Time	lemmatize	0:00:00:00.104	104936100
Time	mkConstEmbeddings	0:00:00:00.372	372005998
Time	parse	0:00:02:33.843	153843165054
Time	recognizedNamedEntities	0:00:00:16.321	16321784881
Time	removeAttachment	0:00:00:00.000	466615
Time	resolveCoreference	0:00:00:00.000	141201
Time	srl	0:00:01:44.776	104776777672
Time	tagPartsOfSpeech	0:00:00:16.957	16957636814
```

Thanks!

MihaiSurdeanu:
Also, I think this branch should be merged in glovecrash4, and the result merged in master. This is much better than the current master. The parser slowdown is caused by the fact that I use 2 fat biLSTMs there. I will work on slimmer models in the future.

kwalcock:
Yes, those are the best times unless you need the intermediate lap times.

PR #475, opened 2021-05-01
Remove .toMap in ConstEmbeddingsglove and debug output

kwalcock:


MihaiSurdeanu:
No need to wait for the tests here.

PR #476, opened 2021-05-01
Optional "improvements" to glove code

kwalcock:
Changes like these were made in the process of debugging so that I could understand what was going on.  They are optional and I don't expect all of them to pass muster.  They are somewhat independent and in order to be cherry picked.


MihaiSurdeanu:
Looks nice, but there are some conflicts.

kwalcock:
This was for glovecrash3.  It will eventually be revisited for use with glovecrash4.

PR #478, opened 2021-05-05
Fix tests to extent possible

kwalcock:
Here are slight updates to tests which will at least allow us to track syntax changes as the code develops, since the tests aren't commented out anymore.  Objects (companion objects or static classes) with state are not especially conducive to testing.  As they accumulate, it becomes more important that they be handled better.


MihaiSurdeanu:
Thanks!

PR #482, opened 2021-05-09
Use lookup parameters for embeddings, but only on a per document basis

kwalcock:
Only words in the document are included, not the entire vocabulary.


kwalcock:
I am going to skip most every "improvement" made to glovecrash3 and sneak them in at a later date, after this version survives longer.  However, there is one change that I think would help it a lot and it's about to be documented.

kwalcock (commented):
After running for approximately 3 hours, this version (actually #483) will crash with memory problems:
```
10:56:32.293 [ScalaTest-run-running-TestParallel] DEBUG org.clulab.dynet.Metal - Loading MTL model from org/clulab/processors/clu/mtl-en-ner-avg-e10e12e8e3e2...

Memory pool info for each devices:
 Device CPU - FOR Memory 128MB, BACK Memory 128MB, PARAM Memory 35136MB, SCRATCH Memory 128MB.
CPU memory allocation failed n=16777216 align=32


CPU memory allocation failed
java.lang.RuntimeException: CPU memory allocation failed
	at edu.cmu.dynet.internal.dynet_swigJNI.ParameterCollection_add_lookup_parameters__SWIG_0(Native Method)
	at edu.cmu.dynet.internal.ParameterCollection.add_lookup_parameters(ParameterCollection.java:63)
	at edu.cmu.dynet.ParameterCollection.addLookupParameters(ParameterCollection.scala:24)
	at org.clulab.dynet.EmbeddingLayer$.load(EmbeddingLayer.scala:306)
	at org.clulab.dynet.Layers$.loadX2i(Layers.scala:176)
```

MihaiSurdeanu:
I think this is worth merging...
This OOM is of course problematic. I wonder if DyNet is NOT freeing its ParameterCollection? Normal usage assumes that they survive model training, so I wonder if they just keep them around?

kwalcock:
The PARAM Memory 35136MB might be a clue.  It shouldn't be needing 35GB for anything.

kwalcock:
Java knows how much memory it has available and collecting garbage long before that memory is exhausted may not be efficient.  I don't think it knows how much C++ memory that DyNet is using up.  It could be a lot more than Java.  I wonder if Java is not being serious about collecting the garbage before the C++ memory usage causes problems.  It might be that reducing the memory that Java can use will force it to clean up before C++ is an issue.  That would be a tough calculation to make, though.

At the Java level all the objects have, in addition to finalize, a delete method which is used to free the C++ memory.  Finalize calls it and it can be called multiple times because it remembers it has been called.  Some manual garbage collecting can be performed so that each lightweight Java object dumps its heavier C++ memory as soon as processors knows the object won't be used anymore, but before the GC gets around to doing it.  It almost seems like what delete was designed to do.  Unfortunately, that Java object is usually not exposed through the Scala object.  However, I could add it.  The ParameterCollection, LookupParameter, and FloatVector that are being used in ConstEmbeddingsGlove and that ConstEmbeddingParameters would be really good candidates for this.

MihaiSurdeanu:
I agree. This makes perfect sense.
In CluProcessor, this is the place to delete stuff: 
https://github.com/clulab/processors/blob/glovecrash4/main/src/main/scala/org/clulab/processors/clu/CluProcessor.scala#L119

There are other places in Metal, but I can handle those if I see an example.

PR #483, opened 2021-05-09
Use just one FloatFector to transfer embeddings

kwalcock:


kwalcock:
Fector->Vector.  My fingers seem to spell differently than my eyes.  They seem not to take voicing into account.  I will consult a linguist.

MihaiSurdeanu:
This makes sense. Thanks!

I suspect the `new FloatVector` is possibly not thread safe in DyNet?

PR #486, opened 2021-05-12
Concept discovery and ranking

BeckySharp:
Moved code from hackathon, incorporated suggestions from @bethard (though I did stick some things on extra lines to make them more readable for me -- sorry!)

I didn't move the App or the ConceptSource, please see other thread.

I'm not 100% in love with what I've exposed in terms of functionality, but maybe it's sufficient for getting started and we can refine the function apis as we go?

FYI @reynoldsm88

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 # Changes
-+ **8.3.3** - Added concept discovery package in `openie`.
++ **8.3.4** - Added concept discovery package in `openie`.
 + **8.3.3** - Handle stopwords in processors, port TagSet and FileUtils from eidos
 + **8.3.3** - Allow KBs to be provided by more generic Standard- and OverrideKbSources.
 + **8.3.3** - Use per sentence lookup parameters for embeddings
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 # Changes
++ **8.3.3** - Added concept discovery package in `openie`.
 + **8.3.3** - Handle stopwords in processors, port TagSet and FileUtils from eidos
 + **8.3.3** - Allow KBs to be provided by more generic Standard- and OverrideKbSources.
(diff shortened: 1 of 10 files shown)


MihaiSurdeanu (approved):

kwalcock:
@BeckySharp thinks this shouldn't go into the 8.3.3 update, so I should probably take what is in master now and publish that.  I updated the one stress test that had been creating multiple processors and exhausting the memory in about 3 hours so that it reuses the same processor like eidos does.  It has been going about 7 hours now.  It may be that the garbage collection doesn't work well with old processors, which would be something to fix but not urgently.

MihaiSurdeanu:
Nice. What do you mean by "garbage collection doesn't work well with old processors"?

Also, agree on the 8.3.3 release.

kwalcock:
meaning that the garbage collector may not be collecting the unreferenced processors before their C++ memory use becomes a problem.

BeckySharp:
> this shouldn't go into the 8.3.3 update

right, at the meeting with two-six today we confirmed moving the concept discovery code to a sep repo for now that depends on processors, and when it's more stable, then moving it in. thanks all for your patience in the meantime and helping me figure out where things should live!

kwalcock:
BTW the test that I was running worked for around 20 hours before I had to get work done myself and shut it off.  Version 8.3.3 was published.  I will then merge this soon then.

MihaiSurdeanu:
Nice!


On May 16, 2021 at 9:40:52 AM, Keith Alcock ***@***.***)
wrote:

BTW the test that I was running worked for around 20 hours before I had to
get work done myself and shut it off. Version 8.3.3 was published. I will
then merge this soon then.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/clulab/processors/pull/486#issuecomment-841842176>, or
unsubscribe

PR #487, opened 2021-05-17
Revert "Concept discovery and ranking"

kwalcock:
Reverts clulab/processors#486


kwalcock:
This purports to undo Keith's overhasty merging of the discovery code.  Sorry for the mix-up.

PR #490, opened 2021-05-20
Show memory exhaustion

kwalcock:


MihaiSurdeanu:
Nice!

kwalcock:
The App InfiniteParallelProcessorExample is a little convoluted because it is based on other programs with a long history.  If reuseProcessor is false (and I didn't change any behavior by rearranging things this morning), the program runs out of memory in approximately 2 hours 50 minutes and has done so three times in a row.  If reuseProcessor is true, then it has run for 20 hours or more (at least once).  The Utils.startup() and shutdown() are supposed to detect memory leaks (if the loop is cut short).  It hasn't worked yet, probably because I need a debug version of FatDynet, but that will be sorted out soon.  The difference is only in that true/false

PR #493, opened 2021-05-22
Use kryo for serialization

kwalcock:


MihaiSurdeanu:
Nice! But tests are failing.

kwalcock:
Passing now

MihaiSurdeanu:
Nice! Merge?
Did you release the artifactory with the binary embeddings?

kwalcock:
This uses the kryo version of glove, which is deployed.  The text version is no longer included as a dependency.

PR #495, opened 2021-05-31
Add some manual memory management by closing DyNet objects when finished

kwalcock:
To CluProcessor add a BeforeAndAfter to manage the temporary attachment.  After removal of the attachment, close the DyNet objects so that C++ memory is freed right away.  Update to new version of FatDynet which includes the close functionality.


MihaiSurdeanu:
Nice, thanks!

kwalcock:
Is it OK to update the CHANGES file and then release so that it can be used with Eidos?

MihaiSurdeanu:
Please do!

PR #497, opened 2021-05-31
Add some documentation of resources and changes

kwalcock:


kwalcock:
This is in preparation of releasing a version 8.3.5 just before other changes are added to master.  In general, the resources we're storing in Artifactory aren't well documented and the procedures to get them there won't be repeatable as soon as my laptop breaks.  However, improving that situation is a separate task which shouldn't slow this down.

MihaiSurdeanu:
Thank you @kwalcock!

PR #498, opened 2021-06-02
added String API...

BeckySharp:
to make the `toDocument` API more flexible

### DIFF
--- a/corenlp/src/main/scala/org/clulab/processors/examples/ProcessorFileExample.scala
+++ b/corenlp/src/main/scala/org/clulab/processors/examples/ProcessorFileExample.scala
@@ -1,6 +1,7 @@
 package org.clulab.processors.examples
 
 import org.clulab.processors.fastnlp.FastNLPProcessor
+import org.clulab.utils.FileUtils
 
 import scala.io.Source
 
@@ -13,7 +14,7 @@ object ProcessorFileExample {
   val proc = new FastNLPProcessor()
 
   def main(args:Array[String]) {
-    val text = Source.fromFile(args(0), "UTF-8").getLines().mkString(" ")
+    val text = FileUtils.getTextFromFile(args(0))
     proc.annotate(text)
   }
 }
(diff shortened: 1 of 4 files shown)


BeckySharp:
idk when the next release is, any chance this would be in it soon?

MihaiSurdeanu (approved):

MihaiSurdeanu:
@kwalcock : any ETA on the latest changes?
Actually, maybe we should release 8.3.6 now. I am cooking some bigger changes for 8.4.0, and I would not rush them.

kwalcock:
The next release is just pending someone saying the word.  (See #496.)  I'm working on the parallelism but it will take some time and may be invasive.

In the meantime, processors has some FileUtils which can get text straight from a file without the open/close.  It might even insist on an encoding which we should use for consistency.  I can add to this PR if nobody minds.

MihaiSurdeanu:
Thanks!

Please add the change and then release.


On June 2, 2021 at 10:22:23 PM, Keith Alcock ***@***.***)
wrote:

The next release is just pending someone saying the word. (See #496
<https://github.com/clulab/processors/pull/496>.) I'm working on the
parallelism but it will take some time and may be invasive.

In the meantime, processors has some FileUtils which can get text straight
from a file without the open/close. It might even insist on an encoding
which we should use for consistency. I can add to this PR if nobody minds.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/clulab/processors/pull/498#issuecomment-853322330>, or
unsubscribe

BeckySharp:
> The next release is just pending someone saying the word. (See #496.) I'm working on the parallelism but it will take some time and may be invasive.
> 
> In the meantime, processors has some FileUtils which can get text straight from a file without the open/close. It might even insist on an encoding which we should use for consistency. I can add to this PR if nobody minds.

thanks! +1 for making the code more consistent with itself!

kwalcock:
Just for the record, some of the old code said
```scala
val contents = source.getLines.mkString
```
`getLines` will remove the EOLs and then mkString defaults to "" when combining, so this is essentially running everything into a single line.  I'm not sure whether that's valid for all JSON and it's probably not what was intended.  I'm going to replace it with
```scala
val contents = source.mkString
```
which preserves line endings and the whitespace that separates tokens, and hope it doesn't cause problems that the tests don't catch.

kwalcock:
This was published last night as 8.3.6 and should be available as a library dependency.

PR #499, opened 2021-06-02
Get text from files more consistently

kwalcock:


kwalcock:
BTW the
```scala
Source.fromFile(args(0), "UTF-8").getLines().mkString(" ")
```
converts EOL characters to spaces.  It seems better not to fool with the characters originally contained in the file.  Analyzer.annotate will have to deal with them.

PR #504, opened 2021-06-11
get Mention.text from raw not words

BeckySharp:
There was an issue I ran into using odin, where the mention.text didn't match the original (specifically, a span like "I'm going to ..." became "Iam doing to..." because of the way the mention.text was created.  This change creates the mention text from the raw.

if a diff functionality is desired, please let me know and I can adjust.

fyi @marcovzla

### DIFF
--- a/main/src/main/scala/org/clulab/odin/Mention.scala
+++ b/main/src/main/scala/org/clulab/odin/Mention.scala
@@ -87,6 +87,9 @@ trait Mention extends Equals with Ordered[Mention] with Serializable {
   /** returns true if the StringMatcher matches any of the mention labels */
   def matches(matcher: StringMatcher): Boolean = labels exists matcher.matches
 
+  /** returns all raw (original, no processing applied) tokens in mention */
+  def raw: Seq[String] = sentenceObj.raw.slice(start, end)
+
   /** returns all tokens in mention */
   def words: Seq[String] = sentenceObj.words.slice(start, end)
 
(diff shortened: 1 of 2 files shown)


MihaiSurdeanu (approved):
Yes, that's the correct way of doing it.
Please merge. And release? @kwalcock ?

main/src/main/scala/org/clulab/odin/Mention.scala
```diff
@@ -140,10 +143,10 @@ trait Mention extends Equals with Ordered[Mention] with Serializable {
     case Some(txt) => txt.slice(startOffset, endOffset)
     case None =>
       // try to reconstruct the sentence using the character offsets
-      val bits = words.head +: tokenInterval.drop(1).map { i =>
+      val bits = raw.head +: tokenInterval.drop(1).map { i =>
```
kwalcock, line 146: It occurs to me that the partner to head is tail.  Since we are assuming a head, tail should be available, so I wonder if tokenInterval.tail would be OK.  It might otherwise be take(1) and drop(1).

BeckySharp: this is rhetorical, right?

BeckySharp:
thanks!

main/src/main/scala/org/clulab/odin/Mention.scala
kwalcock: No, actually.  It's certainly trivial, but I didn't want to change it without a heads up.

BeckySharp:
wok then i can look at the proposed change, this mechanism for building the
text predates me

kwalcock:
It was added to this other tiny change: https://github.com/clulab/processors/pull/508.  I should have gone about it differently :-(

BeckySharp:
no way, @kwalcock , you’re awesome and don’t you forget it!

PR #506, opened 2021-06-12
don't print same Label twice...

### LINKED ISSUE #505: Only print arg label once if same label multiple times
in the RuleReader autodoc for extraction view

BeckySharp:
... in arg table for aggregated view

(not like an urgent important change, but does make it prettier/better)

closes #505


MihaiSurdeanu (approved):

adarshp:
Was just about to post this when I saw you already had an issue for it. For posterity, I'll put a screenshot here of the problem that Becky is addressing with this PR: 

image

MihaiSurdeanu:
Thank you!

PR #513, opened 2021-06-22
Added more numeric date formats with symbol

vanh17:
@MihaiSurdeanu 

I added multiple more tests for all the rules we discussed. 

One more thing I want to ask is that I tried to add 4 more rules as follows:

1) MM/YYYY
2) YYYY/MM
3) YY/MM/DD
4) YY/MM

These rules come from the STUtime Stanford repository. I follow this procedure:
1. add to dates.yml, for each rule, I create the action mkDateMention corresponding.
2. For each mkDateMention action , I create a mkDateMention function inside the NumericActions.scala
3. For each mkDateMention in NumericActions.scala, I create a toDateMention function and a parser.

However, it looks like these new rules are not recognized by the odin. Could you have a look at one rule, let's say yyyy-mm, and go through this process to see if it is correct? I also created tests for that rule, which you can uncomment in the TestDateRecognition.scala.

### DIFF
--- a/main/src/main/resources/org/clulab/numeric/dates.yml
+++ b/main/src/main/resources/org/clulab/numeric/dates.yml
@@ -53,7 +53,7 @@ rules:
     pattern: |
       [word=/^(1\d\d\d|20\d\d)(:|\/|\-)(0?[1-9]|1[012])(:|\/|\-)(0?[1-9]|[12]\d|3[01])$/]
 
-  # Rule for DD-MM-YYYY (accepts -, :, / as separators)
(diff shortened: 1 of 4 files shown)


MihaiSurdeanu:
Thanks! These superficially seem fine. I'll merge, and then try to debug in my branch.

MihaiSurdeanu:
@vanh17 : See this commit to see the changes I had to make for your new tests to pass:
https://github.com/clulab/processors/commit/dbbae97b92368e4398204d1043d6b7fc025a65e7

These are all committed to `numeric_processor`, so no need to do anything. Just FYI.

Issue #492, opened 2021-05-21
Inefficient code


kwalcock:
This is a note to myself to look into making this, line 75 of org.clulab.openie.entities.RuleBasedentityFinder,  more efficient:

```scala
      distinctEntities.filter{ m => stateFromAvoid.mentionsFor(m.sentence, m.tokenInterval, avoidLabel).isEmpty }
```

In order to answer `isEmpty`, the actual values do not need to be calculated in what looks like a significantly expensive operation.

kwalcock:
This is addressed with #515.

PR #515, opened 2021-06-23
Replace collection.filter.isEmpty with !collection.exists

kwalcock:
This is for RuleBasedEntityFinders with supporting methods added to State


kwalcock:
See #492

kwalcock:
This changes some very old code by @myedibleenso, so perhaps he can better recognize whether it might now be broken.

MihaiSurdeanu (approved):

MihaiSurdeanu:
Thanks @kwalcock !

BeckySharp:
@myedibleenso can still check but I think it's safe changes

PR #517, opened 2021-06-28
Add more intensive SUTimes tests and units of Measurement

vanh17:
@MihaiSurdeanu, I added more intensive SUTimes tests for the rest of the grammar. Looks like we passed all of them. Please have a look at those tests and let me know if more you want to ask. Here are my 2 feedback:
1. We can add more grammars to capture winter 2021 as 2021-WI or similar. This can be extended to capture wet season 2021 as 2021-WS like what we read back then in SV documents.
2. I also added 3-4 tests under your TODOs and Marco's. Please let me know if you need help to implement these two TODOs. I think it is best for me to handle the hold things so I can learn more. For example, it might be more efficient if someone asks for help with the numeric processor, all of those requests can go straight to me and not have to sit around waiting.
3. Added complete units of measurements for length, area, volume, and time in the MEASUREMENTS.tsv. Please let me know if you want to change the format as we preliminarily discussed in last week meeting.


MihaiSurdeanu:
Looks good. Thanks!

PR #518, opened 2021-07-01
Sort roots before serialization

kwalcock:
Also update example program


kwalcock:
This was causing failures in regression tests needed for grouping of sentences and would also affect parallel operation.

MihaiSurdeanu (approved):

MihaiSurdeanu:
Thanks @kwalcock !

PR #519, opened 2021-07-03
refactored to make more editable by downstream users if desired

BeckySharp:
Basically there are three entry points for each "view" -- one to get the schema objects, in case you want to mess with the format a lot externally (god help you), one for the string contents of the md file, and one to actually export a md file to a specific path.

@adarshp I also added a "minimal" boolean (to the `toMarkdown` method on the ExtractionSchema as well as the entry point api for extraction schemas, seemed not relevant to the rule schemas) that may do most of what you want, but if not you can always get the objects and write your own code.

### DIFF
--- a/main/src/main/scala/org/clulab/odin/impl/MarkdownGeneration.scala
+++ b/main/src/main/scala/org/clulab/odin/impl/MarkdownGeneration.scala
@@ -1,8 +1,156 @@
 package org.clulab.odin.impl
 
+import org.clulab.odin.impl.MarkdownGeneration._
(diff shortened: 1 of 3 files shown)


kwalcock:
LGTM.  Thanks for closing the files.  I wonder if there will eventually need to be two larger templates that show an overview, like
```scala
val template = """
  | ${preamble}
  | ${table}
  |
  | ${args}
  |
  | &nbsp;
  |"""
```
Some other time maybe.

MihaiSurdeanu:
LGTM too.

adarshp:
Thank you @BeckySharp !

PR #523, opened 2021-07-07
Fix roots of DirectedGraph in CluProcessor

kwalcock:


MihaiSurdeanu:
Nice. Thanks @kwalcock !

kwalcock:
Later in [CluProcessor.srl](https://github.com/clulab/processors/blob/c05ef7fa10b5303fc564f0fa929ee1581117bee1/main/src/main/scala/org/clulab/processors/clu/CluProcessor.scala#L424), the semanticRoles are converted into `enhancedRoles with ToEnhancedSemanticRoles.generateEnhancedSemanticRoles` and there `rolesIndex.toDirectedGraph` is called.  By that time, the roots have changed and the original roots of `rolesIndex` cannot be reused.  They violate the contract.  This is at https://github.com/clulab/processors/blob/c05ef7fa10b5303fc564f0fa929ee1581117bee1/main/src/main/scala/org/clulab/struct/DirectedGraphIndex.scala#L96.

MihaiSurdeanu:
Agreed. The roots for the enhanced roles should be adjusted accordingly.

PR #527, opened 2021-07-10
Fix roots of DirectedGraph

kwalcock:


kwalcock:
I hope to be using this, possibly as a publishLocal, to be reading the many documents starting Saturday.

MihaiSurdeanu (approved):

MihaiSurdeanu:
@kwalcock : this looks great, thanks!
Before merging, would it be possible to do a quick runtime analysis of CluProcessor, before and after this change? I think the runtime overhead of computing the roots should be small because these graphs are small, but I'd like some empirical proof that that's the case.
Thanks!

kwalcock:
Here are some numbers, but my test isn't a very good one.  It's the same test used elsewhere that annotates and now serializes 10 documents 10 times.  The places in the code where the roots are (mis)calculated have changed significantly.  I just added lazy to the value, because the roots are not very often needed.  The roots are still serialized in processors, just for compatibility, and it forces the lazy evaluation, so that's what I timed.  It may be taking slightly longer, but it's difficult to tell given the variability.  In the old version the roots are calculated outside of serialization so that the time would be accounted for there.  It doesn't look like there is a problem.

reroot

| total | serialization |
| ---  | ---  |
| 0:00:05:52.748 | 0:00:00:00.405 |
| 0:00:06:21.538 | 0:00:00:00.526 |
| 0:00:06:29.286 | 0:00:00:00.309 |


master

| total | serialization |
| ---  | ---  |
| 0:00:05:57.029 | 0:00:00:00.352 |
| 0:00:05:49.333 | 0:00:00:00.340 |
| 0:00:06:20.383 | 0:00:00:00.155 |

MihaiSurdeanu:
Thanks! Yes, Ok to merge!

kwalcock:
This branch, locally published, was used to read almost 44,000 documents and did not appear to cause any problems.  As soon as the test passes again, I'll publish.

MihaiSurdeanu:
Nice. Thanks!

PR #535, opened 2021-08-04
Added Mention.norm

MihaiSurdeanu:


main/src/main/scala/org/clulab/odin/Mention.scala
```diff
@@ -102,6 +102,9 @@ trait Mention extends Equals with Ordered[Mention] with Serializable {
   /** returns all entities in mention */
   def entities: Option[Seq[String]] = sentenceObj.entities.map(_.slice(start, end))
 
+  /** returns the norm value of this mention, if available */
+  def norm: Option[String] = sentenceObj.norms.map(x => x(start))
```
BeckySharp, line 106: @MihaiSurdeanu maybe none of my business but why only the start and not the whole slice?

MihaiSurdeanu:
Because the norm of an entity is a single value that is simply replicated in all the tokens of the entity. For example, for the DATE entity "January 1st, 2000" its norm is the single String "2000-01-01".

BeckySharp:
> Because the norm of an entity is a single value that is simply replicated in all the tokens of the entity. For example, for the DATE entity "January 1st, 2000" its norm is the single String "2000-01-01".

@MihaiSurdeanu but what if a mention spans multiple tokens which have diff norms, e.g., a date and a price.  The mention is not necessarily atomic in terms of norm.

MihaiSurdeanu:
The convention I used, similar the corenlp, is that each entity had a
single norm. If it's a nested entity, the norm should aggregate all content.

BeckySharp:
> The convention I used, similar the corenlp, is that each entity had a single norm. If it's a nested entity, the norm should aggregate all content.

@MihaiSurdeanu sorry to keep pushing it's just that I think you're wrong.  Mentions aren't necessarily entities, and you're making a change in general purpose code that doesn't hold for all usages.  An event with several arguments shouldn't have only the norm of the first token, etc.  also, even with an entity-style mention, the norms don't aggregate all content, since the norm doesn't know a priori what the span of the Mention will be and so what would need to be aggregated.  If you want a separate method that does this, that's fine, but I really think this is the wrong implementation here.  At the end of the day, processors is your baby, just trying to help.

BeckySharp:
maybe have `Mention.norm` with this behavior and also `Mention.norms` with the whole slice? Kinda like `label` vs `labels`?

MihaiSurdeanu:
Yes, I think you're right, on both counts. I lean towards changing it to `Mention.norms`, which would be an array of all norms (one per token, from `Sentence.norms`), and let specific apps interpret that.

kwalcock:
It was sounding like I should wait for norms() before releasing.

MihaiSurdeanu:
Yes, thanks!

PR #539, opened 2021-08-08
Hubert meas units

hubert10:
Hi Mihai,

I added some passing & failing tests for measurement units and I would be happy to learn how to write rules to fix them. I have also added few tests on both dates and numbers which I think would be interesting to support them. Thanks

Hubert


MihaiSurdeanu:
Thank you @hubert10!
Quick question: what is the `AVC-MEASUREMENT-UNIT` file?

hubert10:
Hi Mihai, thanks

I put my measurements units in AVC-MEASUREMENT-UNIT file as I was thinking
that Hoang may need to review and validate them before we update the
main MEASUREMENT-UNIT
file. So I think now I can delete it.

Thanks

PR #543, opened 2021-08-11
Hubert added more unit tests on Measurements

hubert10:
Hi Mihai,

Please kindly review this updated **unit tests** on measurements units. You will find some comments labelled **TODO**  for points you may need to decide on how to handle them.

Thanks

Hubert

### DIFF
--- a/main/src/main/resources/org/clulab/numeric/MEASUREMENT-UNIT.tsv
+++ b/main/src/main/resources/org/clulab/numeric/MEASUREMENT-UNIT.tsv
@@ -69,10 +69,14 @@ fl oz               // oz
 pennyweight         // pw
 dwt                 // dwt
 year                // y
+yr                  // y
 years               // y
 nanosecond          // ns
 nanoseconds         // ns
 ns                  // ns
+week                // w
+weeks               // w
+wk                  // w
 day                 // d
 days                // d
 d                   // d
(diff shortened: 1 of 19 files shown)


MihaiSurdeanu:
Thanks @hubert10 !
I'll take over this branch to see what I can do.

MihaiSurdeanu:
Hey @hubert10: I am pushing changing to your branch, which are visible here. This could be a good learning opportunity, to see where to change things in our system.

hubert10:
Alright, sounds good.

Thanks

MihaiSurdeanu:
Hi @hubert10:
A lot of tests that contain "-1" fail because the "-" is not the ASCII character but some other Unicode one. Is this what you get from the PDF converter to text? Or some artifact of your keyboard? If the former, can you please variants for the measurement units to the lexicon that contain both the ASCII "-" and this Unicode char?
Thanks!

MihaiSurdeanu:
For example, this line:
https://github.com/clulab/processors/blob/hubert-meas-units-tests/main/src/test/scala/org/clulab/numeric/TestNumericEntityRecognition.scala#L231
contains a "-" that is not the ASCII character.

MihaiSurdeanu:
Also @hubert10: please pull before working on this branch. I pushed a bunch of changes.

hubert10:
Thanks Mihai,
Yes I will pull before adding anything.

MihaiSurdeanu:
@hubert10 : please don't change the lexicons to add those Unicode chars yet. More soon.

hubert10:
Hi Mihai, Sorry for my late reply.

The way you see them is how they appear in the PDF. I only copy and paste
in my tests. Would you mind checking this on the *Abstract **Section,** Page:
2 (Line 3)  *of this paper
https://hal.archives-ouvertes.fr/hal-00886297/document. You are right they
are not ASCII char. From your comments I used this link to check ASCII
CHAR: https://onlineasciitools.com/validate-ascii

hubert10:
Yes this is well noted. Let me know if I have to change anything.
Thanks

MihaiSurdeanu:
Added @kwalcock.

@hubert10, @kwalcock: I think a better solution than adding these Unicode variations to `MEASUREMENT_UNIT.tsv` would be to preprocess the text files, and normalize some key Unicode characters (like these hyphens) to their ASCII equivalents. What do you think @kwalcock?

What this means for us is:
1. We need a simple pre-processing program to normalize these characters. For example, we could use this file to convert Unicode to ASCII: https://github.com/clulab/processors/blob/master/main/src/main/resources/org/clulab/processors/bionlp/unicode_to_ascii.tsv

2. In `habitus`, we need to make sure that `MEASUREMENT_UNIT.tsv` uses ASCII hyphens, and so do all unit tests.

What do you all think?

kwalcock:
This morning I was going to ask whether something like this hadn't been
done before because I have looked at that mapping file for processors or
reach or something.  Should I or someone try to extract and reuse the
related code?  It would probably look familiar to me after a little reading.

MihaiSurdeanu:
That would be great. Thanks @kwalcock !

I think I've done this during Reach days, but I forgot where that code exists (if any).

kwalcock:
Here are some uninformed but opinionated comments, questionably helpful:

* The primarly culprit here is the en dash '\u2013'
* I would avoid copying it into source code.  Lots of editors work on the code and something will mess them up.  I would go so far as to write `val endash = "\u2013"` and `val text = "ha" + endash + "1"`. to be sure.  Eidos has a unit test that looks through the source code for unicode characters and complains.
* The en dash doesn't itself need to be used to double up entries in MEASUREMENT-UNIT.tsv.  The character doesn't appear in the words, so I don't think it will ever be found.

The en dash seems to be converted into a normal dash during annotation.  However, it will be in a different word.  For a normal dash one gets
```
Words:    Average, yield, reached, 7.2, t, ha-1, in, 1999
Entities: O, O, O, B-MEASUREMENT, I-MEASUREMENT, I-MEASUREMENT, O, B-DATE
```
With an endash it becomes
```
Words:    Average, yield, reached, 7.2, t, ha, -, 1, in, 1999
Entities: O, O, O, B-MEASUREMENT, I-MEASUREMENT, I-MEASUREMENT, B-MEASUREMENT, I-MEASUREMENT, I-MEASUREMENT, B-DATE
```
Here `ha-1` has been split into three words and the dash normalized, but the entities have a B-MEASUREMENT at the dash where an I-MEASUREMENT would be very much preferred.  This prevents both units, each with regular dashes in MEASUREMENT-UNIT.tsv from matching:

t ha-1
t ha - 1

It seems like one way to solve it is to convert
`I-MEASUREMENT, B-MEASUREMENT, I-MEASUREMENT` where the `B-MEASUREMENT` is a dash to `I-MEASUREMENT, I-MEASUREMENT, I-MEASUREMENT`, at least if the `I-MEASUREMENTs` are units.

One can probably find style guides that say the use of an en dash as a minus sign is incorrect.

kwalcock:
I'm not at all sure how this related to habitus, though.

MihaiSurdeanu:
I agree that we should not have the Unicode en dash anywhere in our lexicons or source code. @hubert10: can you please convert the tests to use regular dashes?

But @kwalcock: your solution comes too late because the en dash breaks tokenization. I really think in this case (at least for the Habitus use case), we need to preprocess the docs to convert en dashes to ASCII dashes. This may not be a universal solution, but I think it's the right one here.

kwalcock:
OK, I'll find where that `unicode_to_ascii.tsv` table is used and try to apply it earlier.

MihaiSurdeanu:
Thanks!

hubert10:
Yes, I can do that.thank you

kwalcock:
See #545.  I didn't follow the directions.  We're only having problems with one character and only in certain situations and the unicode_to_ascii.tsv is not going to work right.  I made a pull request on this branch with the start of a solution, changing the en dash to a dash before tokenization.  I'd like to see how far that gets in fixing the problem.  More at #545.

hubert10:
>
> Hi @Keith, Should I continue on this:  @hubert10
> <https://github.com/hubert10>: can you please convert the tests to use
> regular dashes?

   Hubert

kwalcock:
@hubert10 Yes, but possibly better with at least some of the code in #545 and with the expectation that it might need to be revised.  I wanted to submit something so that nobody is just waiting, even if it isn't quite right.

MihaiSurdeanu:
Thanks @hubert10 !

There are still some tests failing. I'll have to inspect those before merging.

hubert10:
Mi Mihai, I think I need to replace en dashes even on the measurements
lexicons. This could fix some of the failing tests. should I?

kwalcock:
I may have just done this.  Hope that it didn't tangle things up.

MihaiSurdeanu:
Yes, Keith's branch handles this, no?

hubert10:
Yes it does.

MihaiSurdeanu:
Finally merged. Thank you all!

@vanh17: please regenerate the outputs with the latest master in processors (you have to publishLocal again) and send them to @mithunpaul08. Thanks!

PR #544, opened 2021-08-14
Now parsing numbers prefixed by "+" correctly.

MihaiSurdeanu:


MihaiSurdeanu:
@vanh17: please take a look at the changes to `NumberParser` here, which fixed this bug.

PR #545, opened 2021-08-14
Replace some en dashes in text for parse

kwalcock:


kwalcock:
This PR shouldn't be merged but parts of it might be copied and pasted over to the branch hubert-meas-units-tests.  Copies of all of the tests with an en dash were collected in one spot.  All the texts are filtered for the en dash and sometimes it is converted to a plain dash.  Some problems are indicated in comments.  Several have to do with constructions like 120 kg N ha-1 where that N, P, K, or urea aren't really units.  Others have to do with ranges.  I don't think the problem is really the dash in the tests still failing.

I'm not sure this change should be in processors or some special project like habitus which happens to deal with documents in which this dash is misused.  It may be that habitus doesn't require strictly preserving the original text like processors would.  There's a todo in there to figure out how that should be done, but that should come after seeing if the substitution works.

Anyway there's a first pass.

MihaiSurdeanu:
Awesome. Thanks @kwalcock !

Please merge.

kwalcock:
This will eventually not go well with one of the other branches in which the en dashes were replaced.  I'd (personally) leave them in the tests, but not as literal characters in the file.  They could be written \u2013 for example.  That's what was done here ,mostly.

PR #546, opened 2021-08-16
Added new tests on numbers and range numbers

hubert10:
Hi Mihai,

I added some tests on numbers and range numbers. Thanks

Hubert

### DIFF
--- a/main/src/test/scala/org/clulab/numeric/TestNumberParser.scala
+++ b/main/src/test/scala/org/clulab/numeric/TestNumberParser.scala
@@ -72,18 +72,24 @@ class TestNumberParser extends FlatSpec with Matchers {
 	}
 
 	it should "parse words to numbers" in {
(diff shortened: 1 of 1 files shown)


MihaiSurdeanu:
Thanks! I will probably get to this only tomorrow, due to other obligations...

hubert10:
Hi Mihai,

Thanks for the comments on tests, could you please point out which
dedicated grammar you usually use for parsing range numbers?

Thanks

Hubert

MihaiSurdeanu:
Hi @hubert10,
The grammar for number ranges is in this branch, which has not been merged yet:
https://github.com/clulab/processors/tree/hubert-meas-units-tests

More exactly, here:
https://github.com/clulab/processors/blob/hubert-meas-units-tests/main/src/main/resources/org/clulab/numeric/number-ranges.yml

Once we merge it, which should be soon, we can revisit number ranges.

hubert10:
Understood. Thanks.

Issue #548, opened 2021-09-08
Basic `CluProcessor` annotation fails for relatively simple operation


reynoldsm88:
### Description
I would like to use `CluProcessor` to go through a list of words and lemmatize them. In the past I used the code similar to the code snippet I've provided, however, now this operation is causing an exception related to something with WordEmbeddings during a sanity check of the document.

I'm not sure if this is something I'm doing wrong or if this is a bug.

### Versions
- java : 1.8.0_202
- scala : 2.12.7
- clulab-processors-main : 8.4.2

### Code
The following code results in an exception in my test case...
```scala
val processor = new CluProcessor()
val doc = processor.mkDocument( "counties" )
processor.lemmatize( doc )
```

### Stacktrace
```java.lang.RuntimeException: ERROR: Const embeddings not set!
[info]   at org.clulab.processors.clu.CluProcessor.basicSanityCheck(CluProcessor.scala:520)
[info]   at org.clulab.processors.clu.CluProcessor.lemmatize(CluProcessor.scala:340)
[info]   at com.acme.MyTestSuite.$anonfun$new$1(MyTestSuite.scala:21)
[info]   at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
[info]   at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
[info]   at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:22)
[info]   at org.scalatest.Transformer.apply(Transformer.scala:20)
[info]   at org.scalatest.flatspec.AnyFlatSpecLike$$anon$5.apply(AnyFlatSpecLike.scala:1683)
[info]   at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)
```

MihaiSurdeanu:
Thanks @reynoldsm88 !

It seems to me that one dependency (the one that handles word embeddings) is not configured correctly. Because of its size, we store it on a local artifactory install, which must be added to the list of repos in the build file. Here is one build file that exemplifies this:

https://github.com/clulab/habitus/blob/main/build.sbt

I don't know if @kwalcock wants to add anything to this.

BeckySharp:
@MihaiSurdeanu do we not need `proc.tagPartsOfSpeech(doc)` first?

MihaiSurdeanu:
Not for the English CluProcessor, afaik.

[...]

PR #549, opened 2021-09-10
Make room for other backend (NN) implementations

kwalcock:
Add a layer of abstraction to support a TorchScript version and others

### DIFF
--- a/corenlp/src/main/scala/org/clulab/processors/fastnlp/FastNLPProcessorWithSemanticRoles.scala
+++ b/corenlp/src/main/scala/org/clulab/processors/fastnlp/FastNLPProcessorWithSemanticRoles.scala
@@ -3,6 +3,8 @@ package org.clulab.processors.fastnlp
 import org.clulab.dynet.{ConstEmbeddingsGlove, Utils}
 import org.clulab.processors.Document
 import org.clulab.processors.clu.CluProcessor
+import org.clulab.processors.clu.backend.EmbeddingsAttachment
+import org.clulab.processors.clu.backend.MetalBackend
 import org.clulab.processors.clu.tokenizer.TokenizerStep
 import org.clulab.processors.shallownlp.ShallowNLPProcessor
 import org.clulab.struct.GraphMap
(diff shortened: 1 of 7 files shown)


kwalcock:
Perhaps we can discuss this at the meeting on Monday.

MihaiSurdeanu (approved):
This makes sense. Thanks @kwalcock!

ZhengTang1120 (approved):

kwalcock:
This is a draft of what might eventually be the branch that integrates the PyTorch code into processors, but probably not until we get as far as inference.  Please see also #550 for ongoing developments.

kwalcock:
I'm beginning to think that the document attachment is unnecessary and overly complex.  I *think* it is there so that the interface is properly satisfied with a simple tagPartsOfSpeech(doc) and recognizeNamedEntities(doc) etc. rather than tagPartsOfSpeedh(doc, embeddings) and recognizeNamedEntities(doc, embeddings).  However, calling tagPartsOfSpeech(doc) without going through annotate(doc) will only result in an exception during the basicSanityCheck anyway, so the individual interface methods are not very useful.  I'm thinking that they should be changed as below with annotate calling the version with the extra argument.  This is also related to the problem that Michael Reynolds reported (#548) recently in which the relatively simple example needed to be modified to account for the embeddings.  It probably isn't really necessary.  The predicate attachment is in a similar situation although it is not sanity checked in the same way.

```scala
  override def annotate(doc:Document): Document = {
    // Do this just once and share the result in a parameter instead of document attachment.
    val embeddings = MetalBackend.mkEmbeddings(doc)

    tagPartsOfSpeech(doc, embeddings)
    recognizeNamedEntities(doc, embeddings)
    chunking(doc)
    parse(doc, embeddings)
    lemmatize(doc)
    srl(doc, embeddings)
    resolveCoreference(doc)
    discourse(doc)
    doc
  }

  override def tagPartsOfSpeech(doc: Document) = {
    // In case this is a one-off call, the embeddings have to be created separately.
    val embeddings = MetalBackend.mkEmbeddings(doc)
    tagPartsOfSpeech(doc, embeddings)
  }

  // Normally the embeddings are supplied as when called from annotate().
  def tagPartsOfSpeech(doc:Document, embeddings: ...): Unit = {
...
```

MihaiSurdeanu:
Thanks @kwalcock !

Also, the doc attachment is necessary to store some intermediate state from CoreNLP, which uses CoreNLP data structures, and I wanted to hide it from the API.

PR #552, opened 2021-09-16
Task Parsing Part for Pytorch Implementation

ZhengTang1120:
@MihaiSurdeanu @bethard @kwalcock 
Here is my current code of the MTL in torch. Thanks to Steve, I already fixed few bugs in the code. 

This is just the task manager and file reader part, I will do another pull after I get the NER task implemented.

### DIFF
--- a/main/src/main/python/pytorch/taskManager.py
+++ b/main/src/main/python/pytorch/taskManager.py
@@ -1,0 +1,213 @@
+import random
+import math
+from sequences.columnReader import ColumnReader
+
+TYPE_BASIC = 0
+TYPE_DUAL = 1
+
+class TaskManager:
+
+  def __init__(self, config, seed):
+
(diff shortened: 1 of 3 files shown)


MihaiSurdeanu:
Looks nice so far!

kwalcock (commented):
How in the world can I delete this comment?

main/src/main/python/embeddings/wordEmbeddingMap.py
```diff
@@ -21,15 +21,13 @@ def load(config):
             else:
                 delimiter = " "
             word, *rest = line.rstrip().split(delimiter)
+            word = "<UNK>" if word == "" else word
```
kwalcock, line 24: IF Python is OK using an empty string as a key, this should not be necessary.

PR #558, opened 2021-10-06
load rules from filesystem if we can find them

marcovzla:
With this pr, the `RuleReader` tries to interpret paths as files first and then as resources in the jar. This is useful for reloading rules without rebuilding.

It also adds a `:reload` option to the shell menu that can be implemented downstream.


MihaiSurdeanu:
Thanks @marcovzla !

PR #560, opened 2021-10-06
Make the shell more inheritable

kwalcock:


kwalcock:
This may not be a good idea, but I'm going to merge this myself.  It's not *supposed* to change any functionality in processors.  The interface is used in habitus.

PR #562, opened 2021-10-07
Enable processor reloading in shells

kwalcock:
Account for reloading in shell menus and reload (i.e., create new processors) when requested. 

Nothing here yet accounts for locations of files.  It is likely that the exact same rules will be used upon reload.

### DIFF
--- a/corenlp/src/main/scala/org/clulab/processors/ProcessorJsonShell.scala
+++ b/corenlp/src/main/scala/org/clulab/processors/ProcessorJsonShell.scala
@@ -1,73 +1,20 @@
 package org.clulab.processors
 
-import org.clulab.dynet.Utils
-import org.clulab.processors.clu.CluProcessor
-import org.clulab.processors.corenlp.CoreNLPProcessor
-import org.clulab.processors.fastnlp.FastNLPProcessorWithSemanticRoles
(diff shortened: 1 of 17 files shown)


kwalcock:
Will do.

kwalcock:
Here in in the just pushed commits is a second draft.  I'm not sure how the directory to support reading (reloading) rules from files should be specified.  In NumericEntityRecognizerShell, I just made it a command line argument.  Perhaps there should be a standard place like the current directory specified in a config file somewhere.

For the VariableShell and VariableProcessor in habitus, the same question applies.  There I also wonder which parts of the VariableProcessor should be reloaded.  There is a lexiconNer that might be changed and also an ExtractorEngine which relies on local files.  In VariableProcessor.apply there is a resource_dir which might serve as the hard-coded directory for reading those rules.  That directory only exists during development and would disappear upon deployment.  I'm not sure when the app will be used.

kwalcock:
It looks like I still need to get the LexiconNER to load from files.  Those seem to only load from resources so far.  habitus is wanting that capability.

MihaiSurdeanu (approved):
Nice. Thank you @kwalcock!

Ok to merge.

PR #564, opened 2021-10-20
Added ability to turn off requirement for N/V in base entity and ability to add to avoid state

BeckySharp:
This is to support further customization of the `CustomizableRuleBasedFinder` in another project.

### DIFF
--- a/openie/src/main/scala/org/clulab/openie/entities/CustomizableRuleBasedFinder.scala
+++ b/openie/src/main/scala/org/clulab/openie/entities/CustomizableRuleBasedFinder.scala
@@ -73,7 +73,7 @@ class CustomizableRuleBasedFinder(
    * @param tag the POS tag to consider
    */
   def isValidTag(tag: String): Boolean = {
-    if (ensureBaseTagNounVerb) tagSet.isAnyNoun(tag) || tagSet.isAnyVerb(tag) else true
+    !ensureBaseTagNounVerb || tagSet.isAnyNoun(tag) || tagSet.isAnyVerb(tag)
   }
 
 
--- a/openie/src/test/resources/test.conf
+++ b/openie/src/test/resources/test.conf
@@ -14,6 +14,8 @@ CustomRuleBasedEntityFinder {
     maxHops = 3             // number of traversals in the syntax dependency graph the system can take to expand the entity
     maxLength = 10          // maximum length the entity can be, in terms of tokens
 
+    ensureBaseTagNounVerb = true
+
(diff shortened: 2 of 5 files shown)


kwalcock:
Someone other than me ;-) might say, "Beautiful, but can you prove that it works with a unit test?"  It seems like a reasonable unit of functionality, running once with true and once with false.  I suspect that it isn't especially handy to test, though, and it may not be worth it unless perhaps the change was prompted by a specific case.

Is the leading _ a reminder that the variable is private?  I don't readily find any other variables using one and would rather not start a new convention.  The underscore seems somewhat problematic in Scala because of how it is otherwise used.  In other languages it is probably less of a concern.  Picky, complaining about single characters I am.

It looks like this avoidState is now a class-level variable while before the stateFromAvoid was method-level and it is furthermore a var.  It seems like we're in danger of having threading issues if extract(doc) is called twice in a single instance of the class.  The variable seems to only be accessed outside that method by the very minimal addToAvoidState() which could just as well be defined inside extract(doc) if need be.  Can you double check?  Maybe there's some recursion I don't see.

Thanks for contributing!  I can make some of the changes so that you're not doing even more work for free.

openie/src/main/scala/org/clulab/openie/entities/CustomizableRuleBasedFinder.scala
```diff
@@ -48,34 +52,33 @@ class CustomizableRuleBasedFinder(
...
+      isValidTag(tags(i)) && !stopNER.contains(entities(i))
     }
   }
 
+  /**
+   * Determines if the tag is a noun/verb
+   * @param tag the POS tag to consider
+   */
+  def isValidTag(tag: String): Boolean = {
+    if (ensureBaseTagNounVerb) tagSet.isAnyNoun(tag) || tagSet.isAnyVerb(tag) else true
```
kwalcock, line 78: This might be written as
```scala
!ensureBaseTagNounVerb || tagSet.isAnyNoun(tag) || tagSet.isAnyVerb(tag)
```

kwalcock: That specific line 78, I mean.

BeckySharp: yes, we could write it that way -- I thought this was a little more readable/parseable, but one man's conditional is another mans.... 
I can change if you want (else you can)

PR #570, opened 2021-11-05
Make reading of comments in lexicons more general

kwalcock:


kwalcock:
@EgoLaparra, can you use something like this?  Start by creating a `Source`, possibly using Sourcer.sourceFromResource (https://github.com/clulab/processors/blob/c8dfa85975eea9f6452220b407a4630940d550f1/main/src/main/scala/org/clulab/utils/Sourcer.scala#L15) or Sourcer.sourceFromFilename (https://github.com/clulab/processors/blob/c8dfa85975eea9f6452220b407a4630940d550f1/main/src/main/scala/org/clulab/utils/Sourcer.scala#L28).  Then call CommentedStandardKbSource.read (https://github.com/clulab/processors/blob/a00751bb3dbb1c19f17990413989ad7dee2196e5/main/src/main/scala/org/clulab/sequences/LexiconNERBuilder.scala#L479) with the source and then a function that takes the left side of each line and optional right side after the comment.  There is an example in UnitNormalizer.readNorms (https://github.com/clulab/processors/blob/a00751bb3dbb1c19f17990413989ad7dee2196e5/main/src/main/scala/org/clulab/numeric/UnitNormalizer.scala#L17).

kwalcock:
This needs some looking at by @EgoLappara and/or @MihaiSurdeanu.  I had even forgotten about it myself.  I will very shortly file an issue and discussion around making the seasons NER configurable.

EgoLaparra:
@kwalcock I've merged this into [dev-season-norm](https://github.com/clulab/processors/tree/dev-season-norm) and made the corresponding changes to use this code. Everything seems to work fine.
Since this is already merged into my [dev-season-norm](https://github.com/clulab/processors/tree/dev-season-norm),  maybe we could close this without merging into master.

kwalcock:
That sounds good, @EgoLaparra.  It looks like it saved about 2 lines of code.  I will close without merging, but keep a local copy of the branch just in case.

PR #575, opened 2021-11-30
Do the easy conversion to Scala 2.13 parts

kwalcock:
These shouldn't cause any problems for the earlier Scala versions, at least not theoretically and according to all the tests.


MihaiSurdeanu:
Thanks @kwalcock !
So return value types have to be explicit in function signatures in Scala 2.13?

kwalcock:
It's a little less that and more that an = is required.  There is still type inference.  It is only a warning for 2.13, but then it won't work at all in 3, These are the easy changes that I had started while waiting on something, but they may as well be merged before conflicts accumulate.

kwalcock:
I hope the thanks implied that it should be merged.

PR #577, opened 2021-12-06
Dev season norm

EgoLaparra:
This pull request includes functionality to parse time expressions including seasons.


kwalcock:
I assume someone decided that it was OK to use astronomical seasons rather than meteorological ;-) and that the approximate start and end dates are close enough.

The new code looks pretty good to me.  There's some old code I don't like, but maybe now is not the time to change it.  For the record, it seems like most RuntimeExceptions in org.clulab.numeric.mentions.package.scala are not really exceptions.  Perhaps some methods should return an Option or, even better, an Either with the left side indicating the problem.

I don't see fall as a season, but maybe it causes too many problems.  I do think it is more prevalent than autumn in American English.

MihaiSurdeanu:
LGTM too!
But I thought the idea was to have just generic seasons in SEASONS.tsv? And customize it in habitus?
I am happy to keep it as is, since the specific seasons you added are probably not ambiguous with other NEs. Just wanted to be clear.

kwalcock:
One of the seasons is in a test directory.  Is that the one in question?  I do have a PR with a few related changes pending.

EgoLaparra:
The specific seasons are just in a resource in test. The main SEASONS.tsv only contains "winter", "spring", "summer" and "autumn".

MihaiSurdeanu:
Cool. Thanks!

kwalcock:
@EgoLaparra, was this ready to merge then?

EgoLaparra:
If you agree, yes!

PR #581, opened 2021-12-10
Add support for before/after time expression

EgoLaparra:
This pull request adds support to recognize time expressions of unbounded date ranges. 
It also covers properly time expressions with seasons a not a year (e.g. "since summer").

### DIFF
--- a/main/src/main/scala/org/clulab/numeric/mentions/DateRangeMention.scala
+++ b/main/src/main/scala/org/clulab/numeric/mentions/DateRangeMention.scala
@@ -15,10 +15,11 @@ class DateRangeMention ( labels: Seq[String],
                          var date2Norm: String )
   extends TextBoundMention(labels, tokenInterval, sentence, document, keep, foundBy, attachments) with Norm {
 
-  // Change date1Norm if the date1Norm does not have a year value but date2Norm has a year value
+  // Change date1Norm if the date1Norm is not undefined (i.e., XXXX-XX-XX) and
+  // does not have a year value but date2Norm has a year value
   // This will handle date ranges of form month date and month date of year.
   // Ex: "Sowing dates between May 3 and October 2, 2020" will have a normalized value of 2020-05-03 -- 2020-10-02
-  if (date1Norm contains "XXXX") {
+  if (!date1Norm.equals("XXXX-XX-XX") && date1Norm.contains("XXXX")) {
     if (!(date2Norm contains "XXXX")) {
       // Keep everything from date1Norm except for the year value, which will be copied from date2Norm.
       date1Norm = date2Norm.substring(0,4) + date1Norm.substring(4)
--- a/main/src/main/scala/org/clulab/numeric/mentions/DateRangeMention.scala
+++ b/main/src/main/scala/org/clulab/numeric/mentions/DateRangeMention.scala
@@ -19,8 +19,8 @@ class DateRangeMention ( labels: Seq[String],
   // does not have a year value but date2Norm has a year value
   // This will handle date ranges of form month date and month date of year.
   // Ex: "Sowing dates between May 3 and October 2, 2020" will have a normalized value of 2020-05-03 -- 2020-10-02
-  if (!date1Norm.equals("XXXX-XX-XX") && date1Norm.contains("XXXX")) {
-    if (!(date2Norm contains "XXXX")) {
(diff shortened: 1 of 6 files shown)


main/src/main/scala/org/clulab/numeric/mentions/package.scala
```diff
@@ -369,11 +383,12 @@ package object mentions {
       if(seasonNorm.isEmpty)
         throw new RuntimeException(s"ERROR: could not find argument season in mention ${m.raw.mkString(" ")}!")
```
kwalcock, line 384: It doesn't especially matter in an error string, but I wonder why m.text is not used.  Perhaps it didn't exist at some point.

main/src/main/scala/org/clulab/numeric/mentions/package.scala
```diff
@@ -339,11 +352,12 @@ package object mentions {
...
 
-      val yearNorm = getArgWords("year", m)
-      if(yearNorm.isEmpty)
-        throw new RuntimeException(s"ERROR: could not find argument year in mention ${m.raw.mkString(" ")}!")
-
-      val (_, yearEnd) = seasonNormalizer.adjustYearRange(seasonNorm.get, yearNorm.get)
+      val yearNorm = getArgWords("year", m) match {
+        case year if year.isDefined =>
```
kwalcock, line 356: Is this `case Some(year) =>` as above?

main/src/main/scala/org/clulab/numeric/mentions/DateRangeMention.scala
```diff
@@ -19,8 +19,8 @@ class DateRangeMention ( labels: Seq[String],
   // does not have a year value but date2Norm has a year value
   // This will handle date ranges of form month date and month date of year.
   // Ex: "Sowing dates between May 3 and October 2, 2020" will have a normalized value of 2020-05-03 -- 2020-10-02
-  if (!date1Norm.equals("XXXX-XX-XX") && date1Norm.contains("XXXX")) {
-    if (!(date2Norm contains "XXXX")) {
+  if (!date1Norm.equals("XXXX-XX-XX") && date1Norm.startsWith("XXXX")) {
+    if (!(date2Norm.equals("ref-date") || date2Norm.startsWith("XXXX"))) {
```
kwalcock, line 23: Has someone been writing too much Java ;-) ?  You almost fooled me.  Is this `date1Norm != "XXXX-XX-XX"`?

kwalcock (commented):
It looks really, really good.

main/src/main/scala/org/clulab/numeric/mentions/package.scala
kwalcock: ...or maybe being raw was very important.

EgoLaparra: To be honest, I did not think much about it. I just followed the same code as the functions above.

main/src/main/scala/org/clulab/numeric/mentions/package.scala
EgoLaparra: oops! yes, it should be the same.

main/src/main/scala/org/clulab/numeric/mentions/DateRangeMention.scala
EgoLaparra: 😅 yes, you a right. I'll fix it right now!
By the way, looking at these 2 lines, the DateNorm companion object you were proposing makes even more sense.

MihaiSurdeanu:
LGTM! @kwalcock: please merge when ready.

kwalcock:
Thanks, @EgoLaparra.  I hope that wasn't too onerous.  You do good stuff.

Tests pass, so I'm pressing the button.

MihaiSurdeanu:
Thanks @kwalcock and @EgoLaparra !

@mithunpaul08 : publishing processors locally should give you access to before/after date ranges.

EgoLaparra:
@kwalcock good stuff? Thank goodness you keep an eye on every mess I try to push 😅 

@mithunpaul08 let me know if you run into any problem.

mithunpaul08:
I agree with @egolaparra. @kwalcock is like the universal try catch block of the lab. If it was not for him i can't imagine how many of my accidental mistakes would have been in production already.

PR #588, opened 2022-01-07
Jollyday integration

EgoLaparra:
This PR includes:
- a new lexicon for holidays extracted from the jollyday library.
- a HolidayNormalizer that uses the jollyday library to retrieve the month and day given a holiday mention and a year. If the year is not given, it uses the current year as default.
- a rule to identify time expressions with holidays.

### DIFF
--- a/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
+++ b/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
@@ -1,0 +1,39 @@
+# list of holidays
+Lincoln's Birthday
+Easter Monday
+Washington's Birthday
+Labour day
+César Chávez's Birthday
+Seward's Day
+Christmas Eve
+Mardi Gras
+Battle of Bennington
+Arbor Day
+West Virginia Day
+Election day
+Confederate Memorial Day
+Statehood day
+Truman Day
+Christmas
+Presidents Day
+Prince Jonah Kuhio Kalanianaole Day
+Independence Day
+New Years Eve
+Jefferson Davis Day
+Lee-Jackson Day
+Patriots' Day
+Alaska Day
+Victory Day
+New Year
+Veterans Day
+Columbus Day
+undefined
+Thanksgiving Day
+Martin Luther King, Jr. Day
+Good Friday
+Casimir Pulaski Day
+Memorial Day
+Service Reduction Day
+Kamehameha Day
+Pioneer Day
+Nevada Day
--- a/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
+++ b/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
@@ -2,9 +2,9 @@
 # List of holidays in the United States extracted from the Jollyday api.
 # This list can be obtained with the following snippet:
 #
-#   import de.jollyday.{HolidayManager, HolidayCalendar}
+#   import de.jollyday.{HolidayManager, HolidayCalendar, ManagerParameters}
 #   import collection.JavaConverters._
-#   val holidayManager = HolidayManager.getInstance(HolidayCalendar.UNITED_STATES)
+#   val holidayManager = HolidayManager.getInstance(ManagerParameters.create(HolidayCalendar.UNITED_STATES))
 #   holidayManager.getCalendarHierarchy.getChildren.keySet.asScala.
 #                   map(holidayManager.getHolidays(2021, _).asScala).flatten.map(_.getDescription)
 #
--- a/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
+++ b/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
@@ -1,4 +1,14 @@
-# list of holidays
+#
+# List of holidays in the United States extracted from the Jollyday api.
+# This list can be obtained with the following snippet:
+#
+#   import de.jollyday.{HolidayManager, HolidayCalendar}
+#   import collection.JavaConverters._
+#   val holidayManager = HolidayManager.getInstance(HolidayCalendar.UNITED_STATES)
+#   holidayManager.getCalendarHierarchy.getChildren.keySet.asScala.
+#                   map(holidayManager.getHolidays(2021, _).asScala).flatten.map(_.getDescription)
+#
+#
 Lincoln's Birthday
 Easter Monday
 Washington's Birthday
--- a/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
+++ b/main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
@@ -1,49 +1,132 @@
 #
 # List of holidays in the United States extracted from the Jollyday api.
-# This list can be obtained with the following snippet:
+# Each phrase before // is an alternative name of a holiday and must be in lower case.
+# The comment after // is the canonical form given by Jollyday.
+# This list of canonical forms can be obtained with the following snippet:
 #
 #   import de.jollyday.{HolidayManager, HolidayCalendar, ManagerParameters}
 #   import collection.JavaConverters._
 #   val holidayManager = HolidayManager.getInstance(ManagerParameters.create(HolidayCalendar.UNITED_STATES))
 #   holidayManager.getCalendarHierarchy.getChildren.keySet.asScala.
-#                   map(holidayManager.getHolidays(2021, _).asScala).flatten.map(_.getDescription)
+#                   flatMap(holidayManager.getHolidays(2021, _).asScala).map(_.getDescription)
 #
 #
-Lincoln's Birthday
-Easter Monday
-Washington's Birthday
-Labour day
-César Chávez's Birthday
-Seward's Day
-Christmas Eve
-Mardi Gras
-Battle of Bennington
-Arbor Day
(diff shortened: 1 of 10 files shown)


MihaiSurdeanu:
Thanks @EgoLaparra !
Shouldn't `holidayManager` in `norm()` be a static argument in the object?

EgoLaparra:
@MihaiSurdeanu you are totally right, it should be!

MihaiSurdeanu:
Thanks! I hope the HolidayManager is thread-safe. 
@kwalcock: any other comments before merging?

kwalcock:
I'll be able to look pretty soon.

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
```diff
@@ -0,0 +1,49 @@
...
+New Years Eve
+Jefferson Davis Day
+Lee-Jackson Day
+Patriots' Day
+Alaska Day
+Victory Day
+New Year
+Veterans Day
+Columbus Day
+undefined
```
kwalcock, line 40: This one isn't right.

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
```diff
@@ -0,0 +1,49 @@
...
+Thanksgiving Day
+Martin Luther King, Jr. Day
```
kwalcock, line 42: AKA MLK Day?  Is it worth adding alternatives?

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
```diff
@@ -0,0 +1,49 @@
...
+Washington's Birthday
+Labour day
+César Chávez's Birthday
+Seward's Day
+Christmas Eve
+Mardi Gras
+Battle of Bennington
+Arbor Day
+West Virginia Day
+Election day
```
kwalcock, line 23: I'm not sure of the capitalization throughout the list, some Day and some day.  Will we match OK either way?

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
```diff
@@ -0,0 +1,49 @@
...
+Confederate Memorial Day
+Statehood day
+Truman Day
+Christmas
+Presidents Day
+Prince Jonah Kuhio Kalanianaole Day
+Independence Day
+New Years Eve
```
kwalcock, line 31: I'm sure that apostrophes will enjoy free variation throughout.  I would have to write New Year's Eve.  Will we match ok on 's, s', or just s?

kwalcock: It might not be something we can solve, but perhaps we should have an idea of what happens.  I don't know offhand how the tokenization works in these cases.  We might get a separate 's as a token.

main/src/main/scala/org/clulab/numeric/mentions/package.scala
```diff
@@ -609,6 +609,36 @@ package object mentions {
...
+
+    case m: RelationMention =>
+      val holiday = getArgWords("holiday", m)
+      if(holiday.isEmpty)
+        throw new RuntimeException(s"ERROR: could not find argument holiday in mention ${m.raw.mkString(" ")}!")
+
+      val year = getArgWords("year", m) match {
+        case Some(y) => Some(y)
+        case _ => None
+      }
```
kwalcock, line 623: This looks suspicious.  Isn't it the same as `val year = getArgWords("year", m)`, unless one might be concerned that `null` is returned.  It doesn't look like it can be.

main/src/test/scala/org/clulab/numeric/TestNumericEntityRecognition.scala
```diff
@@ -175,8 +175,14 @@ class TestNumericEntityRecognition extends FlatSpec with Matchers {
...
-    ensure(sentence= "It was the month of January", Interval(3, 6), goldEntity= "DATE", goldNorm= "XXXX-01-XX")
-    ensure(sentence= "It was the month of January, 2020", Interval(3, 8), goldEntity= "DATE", goldNorm= "2020-01-XX")
+    ensure(sentence = "It was the month of January", Interval(3, 6), goldEntity = "DATE", goldNorm = "XXXX-01-XX")
+    ensure(sentence = "It was the month of January, 2020", Interval(3, 8), goldEntity = "DATE", goldNorm = "2020-01-XX")
+  }
+
+  it should "recognize date from holiday" in {
+    ensure(sentence= "Christmas 2016", Interval(0, 2), goldEntity= "DATE", goldNorm= "2016-12-25")
+    ensure(sentence= "Independence day", Interval(0, 2), goldEntity= "DATE", goldNorm= "XXXX-07-04")
+    ensure(sentence= "between Christmas and New Year", Interval(0, 4), goldEntity= "DATE-RANGE", goldNorm= "XXXX-12-25 -- XXXX-01-01")
```
kwalcock, line 185: Could a test with an apostrophe be added, if only to document how it works or doesn't.  New Year's might suffice.

kwalcock (commented):
I added some comments.  Please fix the English language while you are at it ;-)  I'm attaching a time machine.

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
```diff
@@ -0,0 +1,49 @@
...
+#   import collection.JavaConverters._
+#   val holidayManager = HolidayManager.getInstance(ManagerParameters.create(HolidayCalendar.UNITED_STATES))
+#   holidayManager.getCalendarHierarchy.getChildren.keySet.asScala.
+#                   map(holidayManager.getHolidays(2021, _).asScala).flatten.map(_.getDescription)
+#
+#
+Lincoln's Birthday
+Easter Monday
+Washington's Birthday
+Labour day
```
kwalcock, line 15: Spelt "Labor" around here ;-) but maybe we can't correct their list.

kwalcock:
@EgoLaparra, in case notifications aren't otherwise being sent.

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
EgoLaparra: It is definitely not.

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
EgoLaparra: We could add alternatives and use comments as canonical names. That could look to something like:
```
Martin Luther King, Jr. Day // Martin Luther King, Jr. Day
MLK Day // Martin Luther King, Jr. Day
```
What do you think?

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
EgoLaparra: The lexicon is added as case insensitive and everything is lowercased in HolidayNormalizer, so this should not be a problem. Maybe, I should lowercase this file to avoid any confusion.

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
EgoLaparra: We could use the "comment" trick I mention above. We would use their spellings as canonical forms but we could are our own alternatives:
```
Labour day // Labour day
Labor day // Labour day
```

main/src/test/scala/org/clulab/numeric/TestNumericEntityRecognition.scala
EgoLaparra: I should have tested this in more detail, sorry. I will add more tests.

main/src/main/scala/org/clulab/numeric/mentions/package.scala
EgoLaparra: Yes, checking this again is not necessary.

main/src/main/resources/org/clulab/numeric/HOLIDAY.tsv
EgoLaparra: I've checked and I need to separate the 's in the lexicon, e.g. "Lincoln's Birthday" -> "Lincoln 's Birthday". 

"New Year's Eve" is not recognized. I can add apostrophe alternatives to the lexicon and follow the same "comment as canonical name" I propose to include "MLK Day".

kwalcock:
I don't know how important any of the things are, so maybe Mihai will decide.  I was just looking for things that might go wrong.

MihaiSurdeanu:
I think we should add the variations suggested here. Then merge.
Thanks!

EgoLaparra:
I've added a bunch of alternative forms for the holidays and some additional unit tests.

kwalcock:
It looks awe-fully good to me.  Thanks for dealing with our wonderful language and calendar.

EgoLaparra:
Just learning that Mardi Gras is also called Fat Tuesday was worth it.  😆

kwalcock:
I looked up Pancake Day just to make sure you weren't pulling my leg!

MihaiSurdeanu:
Thank you all!

PR #589, opened 2022-01-14
ActionMirror for Scala 3

kwalcock:


kwalcock:
Scala 3 doesn't have the kind of reflection that is used in Odin's ActionMirror.  What it does have doesn't align nicely, so for Scala 3 I propose this design, pretty much reverting to Java reflection.  This code is actually part of a separate branch that is for all Scala 2.13 and Scala 3 changes, but I'm separating it out, particularly so that @marcovzla has a chance to check it over.  I may well have overlooked something.  This PR can be closed without merging afterwards.  The Scala 2 version will stay the same.  This JActionMirror will become the normal ActionMirror for Scala 3.

MihaiSurdeanu:
Thanks @kwalcock for working on this. Transitioning to Scala 2.13 and then 3 is important.
I'll let @marcovzla comment on this.

kwalcock:
This has been incorporated into a different branch.

PR #591, opened 2022-01-21
Experiment with Onnx

kwalcock:


kwalcock:
@ZhengTang1120, this shows how processors' glove embeddings can be accessed.  On my computer it takes 3-4 seconds to load them while parsing the text file takes about 1:05 minutes.  The faster version is pre-normalized.  The text file needs still to be normalized as it is read in.  I hope that I didn't misinform you about that earlier.  It looks like the unknown vector is already normalized in both cases, despite the TODO in the code.  Perhaps the normalization will help the performance.

MihaiSurdeanu:
Thanks Keith!

PR #593, opened 2022-01-24
Dev single month as date

EgoLaparra:
A few changes to recognize dates with single months (e.g. `it was May`) and an updated version of the timenorm test file that includes those cases.


MihaiSurdeanu:
Thanks @EgoLaparra ! Merged.

PR #599, opened 2022-02-24
Add citations

kwalcock:


kwalcock:
This gets "Cite this repository" added to the home page.  That is really intended to provide a single citation, so the references were used for the list.

image

kwalcock:
It can be tested at https://github.com/clulab/processors/tree/kwalcock/citation.

PR #606, opened 2022-03-25
Do not name resolver simply Artifactory

kwalcock:
A side effect of the simple name is a POM file containing

        <repository>
            <id>Artifactory</id>
            <name>Artifactory</name>
            <url>http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release/</url>
            <layout>default</layout>
        </repository>

which may cause clashes with other respositories.  The new name results in

        <repository>
            <id>artifactorycsarizonaedu</id>
            <name>artifactory.cs.arizona.edu</name>
            <url>http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release/</url>
            <layout>default</layout>
        </repository>

which should play better with others.

### DIFF
--- a/main/build.sbt
+++ b/main/build.sbt
@@ -2,11 +2,11 @@ name := "processors-main"
 description := "processors-main"
 
 pomIncludeRepository := { (repo: MavenRepository) =>
-  repo.root.startsWith("http://artifactory.cs.arizona.edu")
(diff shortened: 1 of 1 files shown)


kwalcock:
I will test this as part of double checking the installation documentation.

kwalcock:
Since we are now using clulab instead, this is not longer applicable.

PR #610, opened 2022-04-01
Update documentation on processors page

kwalcock:


MihaiSurdeanu:
Thanks @kwalcock ! Looks great.

kwalcock:
On the installation page I'd make a distinction between installing for use with Maven and for SBT.  Some people might think that both somehow apply at the same time.

MihaiSurdeanu:
Good idea! You want to do it, or should I?

BeckySharp:
+1 for you two, thanks for caring so deeply about the quality of your library and its users!

kwalcock:
I need to describe having Maven get the http artifacts.  Plus the tests failed and need to be looked at.  Eric from the lab was going to help with arranging for https.

kwalcock:
This is working now.  I'd like to save changes to the installation page for a different PR and go ahead and merge this one.

PR #613, opened 2022-04-02
Fix problem if args have no -

kwalcock:


MihaiSurdeanu:
Nice catch, thanks @kwalcock !

kwalcock:
I had to figure out why my errant positional arguments rather than named ones in pdf2txt would hang the program ;-)

PR #617, opened 2022-04-13
Restore case restoring

kwalcock:


kwalcock:
@hubert10, there are some hints in the small changes included in this PR.

The example text is pre-tokenized.
```
The headline of Thomas Friedman 's column in this morning 's New York Times reads...
```
If something else tokenizes it again, either the CluProcessor or the FastNLPProcessor, it will turn that loose 's into ' s.  If they weren't loose and you had Friedman's and morning's without spaces, then these tokenizers turn them into Friedman 's and morning 's with spaces like you want.  Joining the 's to the word before might be a little tricky, though.

There at least two ways to make a Document: proc.mkDocument(text) and proc.mkDocumentFromTokens(tokens).  You don't quite have text, but the tokens in the other option are expected to be organized into sentences and you don't have that, either, at least not in the example.  Is there a way to extract by sentence from the XML and write them to a file with one sentence per line?  Then you could use the tokens directly.

Will the code be staying in this repo?  I looked for it initially in habitus.  We'll figure out how to do the branching shortly.

@MihaiSurdeanu probably also knows all about the tokenizers, of course.  I'm just guessing.

MihaiSurdeanu:
Thanks @kwalcock and @hubert10 !
Yes, @kwalcock is correct: tokenizing the text twice messes it up.

PR #618, opened 2022-04-13
Added text labels for case restoration

hubert10:
Hi @kwalcock , just to confirm that I was using  your updated modifications on generating labels. Thanks --Hubert

### DIFF
--- a/corenlp/src/main/scala/org/clulab/processors/TextLabelToCoNLLU.scala
+++ b/corenlp/src/main/scala/org/clulab/processors/TextLabelToCoNLLU.scala
@@ -1,13 +1,14 @@
 package org.clulab.processors
 
 import java.io.{File, FileFilter, PrintWriter}
-
-import org.clulab.processors.clu.CluProcessor
+import org.clulab.processors.clu.{CluProcessor, GivenConstEmbeddingsAttachment}
 import org.clulab.processors.fastnlp.FastNLPProcessor
-import org.clulab.utils.StringUtils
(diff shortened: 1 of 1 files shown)


kwalcock:
@hubert10, I should have explained #617 better.  Those were suggested changes made to my branch kwalcock/case-restoring that you might want to incorporate into your case-restoring branch.  Since they are destined for your branch, you get to merge or just close, whichever you see fit, probably using the button at GitHub.  It looks like in this #618 you have merged them somehow (maybe manually or partially, because some of the changes in other files aren't there).  I'm not sure that's what you really want to have done, but in any case, I don't especially need the changes sent back to my branch in this PR.  So, I'd be inclined to click on "Close with comment" and let you continue working on your branch.  Does that make sense?

hubert10:
Yes, it makes sense. I would have added a comment not sent a PR. Thanks

kwalcock:
@hubert10, is this then the final code in TextLabelToCoNLLU.scala that was used?  As far as I can tell, this is the only source file that was changed in this branch and it looks like something that should be merged into the main branch.  It you can confirm, then I will get it there.  Thanks.

hubert10:
Hi Keith, this is correct. This is the final code from
TextLabelToCoNLLU.scala that was used. There was a double tokenization
issue that we dealt with and then that was addressed in the text extraction
from xml in the habitus_utils project.  Thanks!

kwalcock:
Thanks, @hubert10.  The change has been moved to #624 and this #618 and #617 will be closed then.  Just holler if that's not right.

PR #622, opened 2022-04-27
Load components in parallel

kwalcock:


kwalcock:
While testing, I got impatient for processors to start up.  In order to annotate, it requires several "components" that can be loaded independently, and thus in parallel.  Here's an implementation.

| Method | Time1 (sec) | Time2 (sec) | Time3 (sec) |
| --- | --: | --: | --: |
| serial | 47.7 | 47.0 | 48.9 |
| parallel | 11.4 | 11.8 | 13.9 |

I'm looking into the Scala 2.11 compilation problem.

MihaiSurdeanu:
This is awesome :)

kwalcock:
The timing tests above were run with Windows.  When the TestEvalTimeNorm is run on Linux, there's a savings of about 12 seconds.  On the Mac, the most I could see was 3 seconds.  On Linux and Mac, overall time for TestEvalTimeNorm is on the order of 100 seconds.  For Windows it is 1000 seconds.  I'm still trying to explain that, but it doesn't seem to be related to this change.

PR #632, opened 2022-05-14
Provide one implementation of ProgressBar

kwalcock:


kwalcock:
I'm not able to see how it looks in Metal, but there is a sample in TestProgressBar.  It doesn't display just right in IntelliJ during testing, but it seems OK in sbt.  I suspect that the logging in Metal would be a problem, so have added the message to the right of the progress bar.  It may be too long.  There are a couple of formatting options, but this is probably the prettiest.

MihaiSurdeanu:
Thanks @kwalcock!
I can't run it now, but shouldn't the `ProgressBar` be created outside of the `for(metaSentence <- sentenceIterator)` loop rather than inside it?
Thanks!

kwalcock:
I think it is.  This is for the progress through the sentences.  A double progress bar, the other for progress through the epochs, didn't seem to work, but I will check again.
```scala
      val progressBar = ProgressBar(s"Epoch #$epoch", sentenceIterator)
      for(metaSentence <- progressBar) {
```
At one point it was
```scala
      for(metaSentence <- ProgressBar(s"Epoch #$epoch", sentenceIterator)) {
```
but that left no access to the setExtraMessage.

MihaiSurdeanu:
Cool, thank you!

kwalcock:
It is possible to have nested progress bars, but there are a couple of disadvantages.  One is that there is a known bug so that multiple progress bars do not work under Windows anyway.  Another is that there is only one reused progress bar for each level of nesting and it means they need to be reset in an inelegant way and in fact need to operate in an inefficient mode that updates too often because of an implementation detail.  Some information will be lost because the details of only the current epoch will be shown along with a summary of all epochs.  Right now you'd get something like that below.  If you know how many epochs to expect, it should be fairly easy to estimate how much time is left.  I'd add a logging message to print out taskManager.maxEpochs just before that loop.

If the number of epochs is less than the number of lines on screen (after that, the scrolling messes things up), then it might be OK to have a progress bar on the top that counts all the epochs and then have multiple progress bars for the individual epochs.  It wouldn't look right under Windows, but it doesn't crash.   It would work more efficiently and be a more tidy implementation.  The ProgressBar would wrap the Range `0 until taskManager.maxEpochs`.  I'm not sure about ` if epochPatience > 0`.  That seems to be a constant that turns the entire thing on or off and something that might better be placed elsewhere.

I don't see a perfect solution with this library.  I'm inclined to stick with what's there and see how well that works first.

```
[possible line for all epochs here above]
Epoch #0 100% │██████████████████████████████████████████████████│ 5/5 (0:00:02 / 0:00:00) 
Epoch #1 100% │██████████████████████████████████████████████████│ 5/5 (0:00:02 / 0:00:00) 
Epoch #2 100% │██████████████████████████████████████████████████│ 5/5 (0:00:02 / 0:00:00) 
Epoch #3 100% │██████████████████████████████████████████████████│ 5/5 (0:00:02 / 0:00:00) 
Epoch #4 100% │██████████████████████████████████████████████████│ 5/5 (0:00:02 / 0:00:00) 
```

MihaiSurdeanu:
Having a bar per epoch is perfectly fine! That's what I had in mind too.

MihaiSurdeanu:
Nice!

MihaiSurdeanu:
Thanks @kwalcock !

PR #634, opened 2022-05-16
Track progress on evaluation

kwalcock:


kwalcock:
It hasn't been exercised, but it is a miniscule amount of code that can be easily changed.

PR #637, opened 2022-05-30
more date and measurement normalization

maxaalexeeva:
- creating numerical ranges for measurements with dashes (tokenized as one token, e.g., 2-5 cm)
- making a range out of year-season combinations (both one- and two-token combos), e.g., 2017WS
  - ATTN: the information about season is not currently stored
- using year-season combos as a regular year for examples with specific dates, e.g., "from August 10 to October 11 in 2017WS"
- adding relevant tests

### DIFF
--- a/main/src/main/resources/org/clulab/numeric/atomic.yml
+++ b/main/src/main/resources/org/clulab/numeric/atomic.yml
@@ -14,7 +14,7 @@ rules:
     priority: ${ rulepriority }
     type: token
     pattern: |
-      [entity=/B-MEASUREMENT-UNIT/] [entity=/I-MEASUREMENT-UNIT/]*
+      [entity=/B-MEASUREMENT-UNIT/ & !word = "DS"] [entity=/I-MEASUREMENT-UNIT/]*
 
   # possible years, from 1ddd to 20dd
   - name: year
--- a/main/src/main/scala/org/clulab/numeric/NumberParser.scala
+++ b/main/src/main/scala/org/clulab/numeric/NumberParser.scala
@@ -17,7 +17,7 @@ object NumberParser {
           var word = w
           // this is to take care of years with season, e.g., 2011DS, when used in dates with days specified
           if (word.endsWith("WS") || word.endsWith("DS")) {
-            word = word.replace("WS", "").replace("DS","")
+            word = word.dropRight(2)
           }
           // lowercase
           word = word.toLowerCase()
--- a/main/src/main/scala/org/clulab/numeric/NumberParser.scala
+++ b/main/src/main/scala/org/clulab/numeric/NumberParser.scala
@@ -14,8 +14,13 @@ object NumberParser {
         None
       case words =>
         val cleanWords = removePlusMinus(words).flatMap { w =>
+          var word = w
+          // this is to take care of years with season, e.g., 2011DS, when used in dates with days specified
+          if (word.endsWith("WS") || word.endsWith("DS")) {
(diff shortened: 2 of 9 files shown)


main/src/main/scala/org/clulab/numeric/mentions/package.scala
```diff
@@ -121,6 +135,34 @@ package object mentions {
...
+
+    case m: RelationMention =>
+      val date1Norm = getArgNorm("date1", m)
+      if(date1Norm.isEmpty)
+        throw new RuntimeException(s"ERROR: could not find argument date1 in mention [${m.raw.mkString(" ")}]!")
+      val date2Norm = getArgNorm("date2", m)
+      if(date2Norm.isEmpty)
+        throw new RuntimeException(s"ERROR: could not find argument date2 in mention [${m.raw.mkString(" ")}]!")
+
+      new DateRangeMention(
```
kwalcock, line 150: I see quite a few instances of "new DateRangeMention(" with lots of similar code.  It might be time to add another constructor to DateRangeMention that takes the old mention and the two date#Norm Strings and calls the primary constructor.  Another option is to write a helper method that looks similar:
```scala
def newDateRangeMention(mention: Mention, date1Norm: String, date2Norm; String): DateRangeMention = {
  new DateRangeMention(
    mention.labels,...
```

kwalcock (commented):
Thanks for understanding these rules so well!

main/src/main/scala/org/clulab/numeric/NumberParser.scala
```diff
@@ -14,8 +14,13 @@ object NumberParser {
         None
       case words =>
         val cleanWords = removePlusMinus(words).flatMap { w =>
+          var word = w
+          // this is to take care of years with season, e.g., 2011DS, when used in dates with days specified
+          if (word.endsWith("WS") || word.endsWith("DS")) {
```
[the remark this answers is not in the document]
maxaalexeeva: Great point! Adjusting that!

main/src/main/resources/org/clulab/numeric/atomic.yml
```diff
@@ -14,7 +14,7 @@ rules:
     priority: ${ rulepriority }
     type: token
     pattern: |
-      [entity=/B-MEASUREMENT-UNIT/] [entity=/I-MEASUREMENT-UNIT/]*
+      [entity=/B-MEASUREMENT-UNIT/ & !word = "DS"] [entity=/I-MEASUREMENT-UNIT/]*
```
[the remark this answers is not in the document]
maxaalexeeva: "d" is found as a unit (days) after lemmatization I think. I am basically making sure capitalized DS is just never captured as a unit. Currently, there is no `w` unit.

main/src/main/scala/org/clulab/numeric/mentions/package.scala
maxaalexeeva: @kwalcock would you check if my last commit is what you meant with this comment?

kwalcock: An apply method on the companion object is just as good or even better.  Thanks.

kwalcock:
In case it isn't clear, I haven't merged because a review from Mihai was also requested.  It seems like that would mean "and" rather than "or".

maxaalexeeva:
@kwalcock, makes sense!

MihaiSurdeanu:
Beautiful. Thank you @maxaalexeeva and @kwalcock !
Ready to merge.

Issue #639, opened 2022-06-09
It should work with new versions of corenlp


kwalcock:
It is failing some tests in TestCoreNLPProcessor and TestFastNLPProcessor.  Details will be pasted below.

kwalcock:
Below are tests for TestCoreNLPProcessor.  Search for "*** FAILED ***"

```
sbt:root> testOnly org.clulab.processors.TestCoreNLPProcessor
[info] compiling 1 Scala source to D:\Users\kwa\Documents\MyData\Projects\clulab\processors-project\processors\main\target\scala-2.12\test-classes ...
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] No tests to run for Test / testOnly
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] No tests to run for main / Test / testOnly
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] No tests to run for openie / Test / testOnly
Picked up _JAVA_OPTIONS: -Xmx20g -Dfile.encoding=UTF-8
[info] TestCoreNLPProcessor:
[info] CoreNLPProcessor
12:33:57.270 [pool-1-thread-1-ScalaTest-running-TestCoreNLPProcessor] INFO edu.stanford.nlp.pipeline.StanfordCoreNLP - Adding annotator pos
12:33:59.829 [pool-1-thread-1-ScalaTest-running-TestCoreNLPProcessor] INFO edu.stanford.nlp.tagger.maxent.MaxentTagger - Loading POS tagger from edu/stanford/nlp/models/pos-tagger/english-left3words-distsim.tagger ... done [2.5 sec].
12:34:00.002 [pool-1-thread-1-ScalaTest-running-TestCoreNLPProcessor] INFO edu.stanford.nlp.pipeline.StanfordCoreNLP - Adding annotator lemma
12:34:00.298 [pool-1-thread-1-ScalaTest-running-TestCoreNLPProcessor] INFO edu.stanford.nlp.pipeline.StanfordCoreNLP - Adding annotator ner
12:34:00.418 [pool-1-thread-1-ScalaTest-running-TestCoreNLPProcessor] INFO edu.stanford.nlp.sequences.SeqClassifierFlags - maxAdditionalKnownLCWords=0
[...]

PR #640, opened 2022-06-09
Make run with StanfordNLP 4.4.0

kwalcock:


kwalcock:
See #639 for errors

kwalcock:
This will be resubmitted via the modernize branch.

Issue #612, opened 2022-04-01
Java 17


kwalcock:
According to reliable sources, it may require in the call to Java

--add-opens java.base/sun.net.[www.protocol.jar](http://www.protocol.jar/)=ALL-UNNAMED

See https://docs.oracle.com/en/java/javase/16/migrate/migrating-jdk-8-later-jdk-releases.html

This command may help diagnose problems related to missing classes:

mvn dependency:build-classpath

kwalcock:
To add new Scala code but still work in Maven, one might need things like this:
```xml
              <plugin>
                <groupId>net.alchim31.maven</groupId>
                <artifactId>scala-maven-plugin</artifactId>
                <version>4.6.1</version>
                <executions>
                    <execution>
                        <id>scala-compile-first</id>
                        <phase>process-resources</phase>
                        <goals>
                            <goal>add-source</goal>
                            <goal>compile</goal>
                        </goals>
                    </execution>
                    <execution>
                        <id>scala-test-compile</id>
                        <phase>process-test-resources</phase>
                        <goals>
                            <goal>testCompile</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <scalaVersion>2.12.15</scalaVersion>
                </configuration>
            </plugin>

```

```xml
        <dependency>
            <groupId>org.scala-lang</groupId>
            <artifactId>scala-library</artifactId>
            <version>2.12.15</version>
        </dependency>
        <dependency>
            <groupId>org.clulab</groupId>
            <artifactId>processors-main_2.12</artifactId>
            <version>8.4.8</version>
        </dependency>
    ...
    <repositories>
        <repository>
            <id>artifactory.cs.arizona</id>
            <url>http://artifactory.cs.arizona.edu:8081/artifactory/sbt-release</url>
        </repository>
    </repositories>
```

MihaiSurdeanu:
Nice @kwalcock !

Should we add a sample pom.xml file somewhere in the documentation?

kwalcock:
This issue is here mostly to preserve Becky's findings after Slack forgets it and until we advance to more recent Java versions.  If I understand it correctly, the XML above is mostly helpful for people wanting to add a few Scala source files to their otherwise Java project built with Maven.  It isn't directly related to processors.  We usually work the other way around with a Scala project built with sbt to which a few Java files are added.  I don't think we need to document for others how this is done.  We don't document gradle, either, for example.

MihaiSurdeanu:
[...]

PR #641, opened 2022-06-10
Allow compilation and run with up to Java 18

kwalcock:
This does not up the version on Stanford NLP, however.  That will be a separate PR after the tests pass and will necessitate incrementing the version number substantially.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ Makefile
 project/
 
 # sbt specific
+.bsp
 dist/*
 target/
 lib_managed/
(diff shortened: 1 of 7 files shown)


kwalcock:
See also #612 for more information.  No special flags are needed for Java now.

MihaiSurdeanu:
Thanks @kwalcock ! Ready to merge, as far as I can tell.

kwalcock:
There may be unintended side effects, but merging now will help us find them before the release to maven.

kwalcock:
@BeckySharp, you might be interested in this PR.  Things like --add-opens java.base/sun.net.[www.protocol.jar](http://www.protocol.jar/)=ALL-UNNAMED should no longer be needed.

Issue #280, opened 2018-10-09
Mention.withAttachment and Mention.withoutAttachment ignore CrossSentenceMention


kwalcock:
These case statements in Mention are incomplete and will result in runtime errors for CrossSentenceMentions.  These should not be case statements at all, but instead methods in each subclass.

```
  def withAttachment(mod: Attachment): Mention = this match {
    case m: TextBoundMention => m.newWithAttachment(mod)
    case m: RelationMention => m.newWithAttachment(mod)
    case m: EventMention => m.newWithAttachment(mod)
  }

  def withoutAttachment(mod: Attachment): Mention = this match {
    case m: TextBoundMention => m.newWithoutAttachment(mod)
    case m: RelationMention => m.newWithoutAttachment(mod)
    case m: EventMention => m.newWithoutAttachment(mod)
  }
```

kwalcock:
There's also no way to copy a CrossSentenceMention.

Issue #277, opened 2018-10-02
Add multiple attachments at a time to a Mention


kwalcock:
We need to be able to avoid code like that below where we are forced to copy entire mentions multiple times in order to add multiple attachments.  Such code has appeared more than once.

```
  // If any hedging was found, store it as an attachment
  def withApplicableHedging(m: Mention, hedgedLemmas: Seq[String]): Mention = {
    var finalMention = m
    for (hl <- hedgedLemmas) {
      finalMention = finalMention.withAttachment(Hedging(hl, None))
    }
    finalMention
  }
```
This should become
```
   val finalMention = m.withAttachments(hedgedLemmas.map(Hedging(_, None)))
```

PR #646, opened 2022-06-17
Add multiple attachments at a time

kwalcock:
This is to address #280


kwalcock:
Make that #277.

kwalcock:
I'm not sure what the problem was, but it seems unrelated to this code that it never called.  I'm writing "suspicious" here in case it needsto be found again.

Issue #316, opened 2019-02-28
Questionable code for matching brackets and braces


kwalcock:
Please see https://github.com/clulab/eidos/issues/300 for details.  Not particularly important, but at least for the record.

PR #647, opened 2022-06-17
Port matchingBrackets from Eidos

kwalcock:
First fail the test then pass it.  More extensive tests are in Eidos.

This is for #316


MihaiSurdeanu:
Thanks @kwalcock !

Issue #188, opened 2017-10-06
Dataset should extend Traversable


myedibleenso:
It would be convenient if [`Dataset`](https://github.com/clulab/processors/blob/54d8e7b96bc0f03a9812e09bd0f360d96c241ad0/main/src/main/scala/org/clulab/learning/Dataset.scala) extended [`Traversable`](http://docs.scala-lang.org/overviews/collections/trait-traversable.html).  This would simplify access to each `Datum` via `foreach`, `map`, etc.

Right now for a `Dataset` (`ds`) I'm using something like `ds.labels.indices.map(data.mkDatum)`.

kwalcock:
@myedibleenso, this was finally done and is currently available in the master branch and will be in the next release!  You would need to change your code to something like `ds.map` to make use of it.

PR #648, opened 2022-06-17
Dataset should extends Traversable (or something similar)

kwalcock:
This addresses #188.  I do not know for sure that idx and row are the same thing.  If they aren't, then the indices could be off or even out of range.  Can someone more familiar give it a quick check?


MihaiSurdeanu:
Yes, `idx` indexes rows in the dataset, so your implementation seems correct.

Issue #642, opened 2022-06-14
Normalization issues


maxaalexeeva:
- [ ] number + preposition of time/place (`in`) mistakenly labeled as measurement:
```
ex. 1. Popular varieties in the wet season were Sahel 202 (65% of farmers) and Sahel 201 (30%), while 60% and 92% grew Sahel 108 in 2012DS and 2013DS, respectively.
ex. 2. released as Sahel 108 in Senegal in 1994
ex. 3 rice yield will increase from 3600 in 2000-2009 to 4500 kg ha-1 in 2090-2099 ( Fig. 4a ) . 
``` 
- [ ] unclear how to normalize the numbers in parens (may need to be addressed directly in habitus):
```
In plots receiving fertilizer, DAP was applied basally (19.3 and 21.5 kg N and P ha-1 )
```
- [ ] pluses are included as B-MEASUREMENT: feature or a bug? (norms are correct, e.g., 100.0 kg/ha for + 100 kg ha-1):
```
das days after sowing, Fert fertilizer treatment, with F1: recommended dose (80 kg N ha−1), i.e., 200 kg ha−1 NPK (15.15.15) at sowing + 100 kg ha−1 urea at 20 das + 50 kg ha−1 urea at 50 das. F2: F1/4 (20 kg N ha−1)
```
- [ ] range not extracted:
```
Ex. 1. Rice grain yield measured at maturity ranged from 2.7 t ha-1 to 7.1 t ha-1 , with an average of 4.8 t ha-1 .
Ex. 2 rice grain yields were between 8.8 t ha-1 and 9.2 t ha-1 ( i.e. about 1 t ha-1 more than in the 1998WS
```
- [ ] how to normalize two separate values with shared unit, e.g., 9 (t/ha) and 10 t/ha below? 
```
Potential yield of all the varieties in the Senegal River delta was estimated at 9 and 10 t / ha in wet and dry seasons , respectively , and potential yield was taken as 8 t / ha for both seasons in the middle valley .
Ex. 2 rice yield will increase from 3600 in 2000-2009 to 4500 kg ha-1 in 2090-2099 ( Fig. 4a ) .
``` 
Probably similar to three values in the next example. Maybe extract as one unit and split with an action? (unsure if it will be easy to differentiate these from ranges---which will need to be stay unsplit):
```
Target yields on average were set to 6.4 , 7.9 , and 7.1 t / ha in 2011WS , 2012DS , and 2013DS , respectively ( Table 1 ) .
```

[...]

PR #649, opened 2022-06-21
more norms

maxaalexeeva:
- addressing some of the issues in https://github.com/clulab/processors/issues/642 
- adding relevant tests
- allowing a negative check in numerical entity tests (e.g., to make sure we did not extract `108.0 in` from `Sahel 108 in Senegal`)

### DIFF
--- a/main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
+++ b/main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
@@ -57,9 +57,7 @@ class NumericActions(seasonNormalizer: SeasonNormalizer) extends Actions {
 
   /** Constructs a MeasurementMention from a token pattern */
   def mkSharedMeasurementMention(mentions: Seq[Mention], state: State): Seq[Mention] = {
-    val mens = convertOneToMany(mentions, toSharedMeasurementMention, "toSharedMeasurementMention")
-    mens.foreach(m => println("MEN: " + m.text))
-    mens
+    convertOneToMany(mentions, toSharedMeasurementMention, "toSharedMeasurementMention")
   }
 
   def mkPercentage(mentions: Seq[Mention], state: State): Seq[Mention] = {
@@ -142,6 +140,11 @@ class NumericActions(seasonNormalizer: SeasonNormalizer) extends Actions {
   /** Constructs a DateRangeMention from a token pattern */
   def mkDateRangeMentionVagueSeason(mentions: Seq[Mention], state: State): Seq[Mention] = {
     convert(mentions, toDateRangeMentionFromVagueSeason, "mkDateRangeMentionVagueSeason")
+  }
+
+  /** Constructs a DateRangeMention from a token pattern */
+  def mkDateRangeMentionOneTokenYearRange(mentions: Seq[Mention], state: State): Seq[Mention] = {
+    convert(mentions, toDateRangeMentionFromOneTokenYearRange, "mkDateRangeMentionOneTokenYearRange")
   }
 
   /** Constructs a DateMention from a token pattern */
--- a/main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
+++ b/main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
@@ -16,7 +16,24 @@ class NumericActions(seasonNormalizer: SeasonNormalizer) extends Actions {
     val convertedMentions = new ArrayBuffer[Mention]()
     for(m <- mentions) {
       try {
-        convertedMentions += converter(m)
+        convertedMentions += converter(m )
+      } catch {
+        case e: Exception =>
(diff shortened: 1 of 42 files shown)


maxaalexeeva:
@kwalcock would you please give me the power to check what jenkins is up to?

main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
```diff
@@ -16,7 +16,24 @@ class NumericActions(seasonNormalizer: SeasonNormalizer) extends Actions {
...
+      } catch {
+        case e: Exception =>
+          // sometimes these conversions fail, mainly on broken texts
+          // let's be robust here: report the error and move on
+          System.err.println(s"WARNING: $converterName conversion failed! Recovering and continuing...")
+          e.printStackTrace()
+      }
+    }
+    convertedMentions
+  }
```
kwalcock, line 29: This isn't necessary, but one way to do it without the buffer is
```scala
    for {
      m <- mentions
      converted <- Try(converter(m))
          .recover { case throwable: Throwable =>
            System.err.println(s"WARNING: $converterName conversion failed! Recovering and continuing...")
            throwable.printStackTrace()
            Seq.empty[Mention]
          }
          .get
    } yield converted
```

main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
```diff
@@ -16,7 +16,24 @@ class NumericActions(seasonNormalizer: SeasonNormalizer) extends Actions {
...
     for(m <- mentions) {
       try {
-        convertedMentions += converter(m)
+        convertedMentions += converter(m )
+      } catch {
+        case e: Exception =>
+          // sometimes these conversions fail, mainly on broken texts
+          // let's be robust here: report the error and move on
+          System.err.println(s"WARNING: $converterName conversion failed! Recovering and continuing...")
+          e.printStackTrace()
```
kwalcock, line 25: Probably all of processors needs to be reviewed for its use of println, System.out, and System.err.  These should probably be logged rather than printed.  That's not your problem.

kwalcock (approved):
LGTM but I wish that meeting the needs of habitus wasn't requiring changes to processors.  It should be possible to adapt locally.  That's not your problem, though.

main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
maxaalexeeva: Makes sense. With processors I basically go the path of fewest changes :)

maxaalexeeva:
Hi @kwalcock, I notice that the version in project is still the old snapshot (8.5.1-SNAPSHOT). Maybe I branched off unpulled master. However, I do not see this difference reflected here under Files changed. Should I merge master into my branch to fix this? I am a little confused.

kwalcock:
AFAICT this branched from master just before 8.5.1 was released.  There are no problems when I merge locally.  We've probably just been working on non-intersecting parts of the code.  I don't think there is need to merge master into this branch without merge conflicts.

maxaalexeeva:
@kwalcock thanks for checking!

maxaalexeeva:
@kwalcock @MihaiSurdeanu fyi, converting this to draft for now. I am going through habitus tests to see what else needs fixed and whether these changes help in habitus, and I keep finding more things to do.

maxaalexeeva:
@MihaiSurdeanu I think I am done with this for now. Please have a look. 

One concern I have is this commit https://github.com/clulab/processors/pull/649/commits/2b935c669f70f8b953949da68f3fbe4ae99a786d 
The intention was to avoid finding `in` in `released as Sahel 108 in Senegal in 1994` as a unit (`inch`). My previous solution of just disallowing units before NNPs did not work because fertilizers in habitus are frequently tagged as NNP. I also can't add a constraint to the rule based on location entity because based on what I can tell from the code, entities like locations are not available at the stage of numerical info extraction. My solution, though, feels to be a bit overkill. Any suggestions? If not, keep my solution or let it go with this example?

MihaiSurdeanu:
Hi @maxaalexeeva,
Yes, that commit is problematic. Would it be simpler to disallow units if the POS tag is "IN" or "TO"? Assuming the POS tagger is correct, this should fix the "in Senegal" bug because "in" there should be a preposition.
I think the question is: does the POS tagger label units as IN in legit cases?
Thanks!

maxaalexeeva:
@MihaiSurdeanu if I remember correctly, in(ch) got tagged as IN as well, but I'll double check tomorrow.

MihaiSurdeanu:
I see...
Then I don't have a better solution. Please merge it as is, @maxaalexeeva or @kwalcock.
Thanks!

maxaalexeeva:
@kwalcock I don't think I have the power to merge, so feel free to merge if you have no further comments!

kwalcock:
Will do.

PR #650, opened 2022-06-21
Avoid possible runtime errors when using AutoCloser

kwalcock:
Eidos was also having runtime reflection errors with newer versions of Java.  I decided that the runtime reflection is too dangerous for clients to use.  They will need to declare inheritance from Java's Closeable.  This may break compilation, but that is so that it doesn't break at runtime.


kwalcock:
It's very strange that an unrelated test failed and I will be on the lookout for a problem.  This PR is being replaced by #651 which has a much better solution.

PR #654, opened 2022-07-08
Prevent crash in Eisner algorithm

kwalcock:


kwalcock:
I'm not certain this is exactly what you had in mind.  It seemed to have fixed the crash without breaking anything else.

MihaiSurdeanu:
Looks beautiful. Thanks @kwalcock!

PR #656, opened 2022-07-18
Do not trounce generic (non-O) NER label with custom

kwalcock:


MihaiSurdeanu:
It should be changed. Thanks!

MihaiSurdeanu:
Sorry @kwalcock, I think I screwed up the description of the task...
The custom NER should overwrite existing labels in two cases:
- They are currently "O". You handle this case now.
- They are a different label, but the current named entity (i.e., the span from the first B- to the last I-) overlaps *exactly* with the span that would be annotated in the lexicon. So, in the unit test above, the B-D and I-D labels should overwrite the B-PER and I-PER because the spans are identical.

Sorry!

kwalcock:
No problem.  There will be an update soon.

MihaiSurdeanu:
Beautiful. Thanks @kwalcock !

PR #658, opened 2022-07-30
2.13 cross compile support

thebrenthaines:
I have used the processors package quite a bit in projects of my own. It has recently become necessary to support Scala 2.13, so this PR does that. It is a relatively big PR, but unit tests work for 2.13, 2.12 and 2.11. I've tried to resist the temptation to make idiomatic or style changes (a few minor ones snuck in).

I've also tried to be conscious of the optimizations that have been part of the codebase. Sometimes it just wasn't reasonable. For instance overriding the HashMap class so you can change the initial capacity relies on functionality that isn't available in 2.13. In the cases where the optimization was not easily supported in 2.13, I opted for eliminating the optimization rather than adding to the version-specific code footprint.

The PR utilizes some adapter classes in version-specific source folders. I've worked to minimize the footprint of the version-specific code.
---
* Added support for 2.13. The main source of trouble was from Scala's updated collections implementation
* Substituted `Map.mapValue` for `Map.map` to assure cross platform integrity
* Addressed deprecation wherever it could be done without breaking backwards builds
* Made explicit calls to `Array.toIndexedSeq`, as the implicit conversion is both deprecated and slow
* Procedure declaration style has been deprecated for a *long* time, I eliminated it wherever I found it.
* Updated dependencies to be more modern wherever 2.13 support was unavailable
* IMO the riskiest change is away from Array.view, which in 2.13 is both deprecated and uses a copy of the array, so some functions were broken (specifically, the `norm` method in `WordEmbeddingsMap`). In this case, I found the best way to update this so we could maintain the same performance characteristics, was to provide a version of `norm` that takes the sub-array indexes and work against the backing array.


kwalcock:
Wow, nice!  It's going to take a lot of time to digest this.  It would have been nice to collaborate.  You couldn't have known, but we had a local branch, [scala2.13](https://github.com/clulab/processors/tree/kwalcock/scala2.13), or two, [scala3](https://github.com/clulab/processors/tree/kwalcock/scala3), already in the works that I just pushed for comparison.  I do not recall right away what was preventing the update.  It's probably time to revisit.

MihaiSurdeanu:
Thank you @thebrenthaines ! We will carefully look at this.

kwalcock:
I've been working on reconciling the two sets of changes.  In many many cases they match and afterwards perhaps @thebrenthaines can look at the few that remain to see if the clulab's version is reasonable and avoid any forking.  I made the mistake of basing the Scala 3 branch off the 2.13 branch, so it would be useful to not switch to this branch so that continuing through to Scala 3 might be easier.  I especially appreciate the changes like

```scala
    for(i <- 0 until avgWeights.size) {
    // becomes
    for(i <- avgWeights.indices) {
```
and many replacements of string addition.

thebrenthaines:
> I've been working on reconciling the two sets of changes. In many many cases they match and afterwards perhaps @thebrenthaines can look at the few that remain to see if the clulab's version is reasonable and avoid any forking. I made the mistake of basing the Scala 3 branch off the 2.13 branch, so it would be useful to not switch to this branch so that continuing through to Scala 3 might be easier. I especially appreciate the changes like
> 
> ```scala
>     for(i <- 0 until avgWeights.size) {
>     // becomes
>     for(i <- avgWeights.indices) {
> ```
> 
> and many replacements of string addition.

I'm happy to take a look when you're ready. Just let me know and I'll jump on it.

I wonder if it would be worthwhile merging / rebasing the clulab 2.13 branch forward to the latest master first? (Maybe you've already done this locally). This PR is my second attempt at making processors compile with 2.13, the first was about 9 months ago and something popped up that made me say "screw it" ;). I can't recall what that was, but I had an unexpectedly straightforward time with the latest master branch, so...

kwalcock:
I did merge master back into the 6-month-old Scala 2.13 branch first thing (and was reminded why it's not a good idea to wait so long) :-(

thebrenthaines:
Yeah. I think we've all re-learned that lesson a few times. :) Let me know when the branch is ready and I'll take a look. I'm happy to help.

kwalcock:
So I'd like to close this PR unmerged and substitute #661 in its place.  Even if the files were exactly the same, the commit history is different and that might eventually be important.  They are different, though, and I diffed all of them to see what I might have missed or should have been done differently.  Below is a summary of some of the things I noticed.  Right now the branch is compiling and testing cleanly, so it's OK to look at the [code](https://github.com/clulab/processors/tree/kwalcock/scala2.13merge) and send feedback.  I'm still going to do some regression and performance tests.

| PR #658 | PR #661 | Notes |
| --- | --- | --- |
| import scala.collection.BufferedIterator | import org.clulab.scala.BufferedIterator | org.clulab.scala is intended for code that smoothes out differences between Scala versions. |
| array.slice | import org.clulab.utils.ArrayView<br>import org.clulab.utils.MutableArrayView | Check performance |
| HasParallelSupport | ThreadUtils | 661 is less elegant, but interfaces better with our other projects, I think. |
| import scala.collection.JavaConverters._ | import scala.jdk.CollectionConverters._ | This may be foreshadowing Scala 3. |
| Stream | import org.clulab.scala.LazyList | Likewise |
| import org.scalatest.flatspec.AnyFlatSpec<br>import org.scalatest.matchers.should.Matchers | import org.scalatest.{FlatSpec, Matchers} | I avoided needing this change by holding back on the scalatest update.  A different round is planned for this.  This fix will be to inherit from Test as in this [example](https://github.com/clulab/processors/blob/kwalcock/scala2.13merge/main/src/test/scala/org/clulab/utils/TestTimer.scala).  |
| toIndexedSeq | sometimes toSeq when possible | Check performance |

thebrenthaines:
> So I'd like to close this PR unmerged and substitute #661 in its place. Even if the files were exactly the same, the commit history is different and that might eventually be important.

Works for me. Is there any work you deferred that you would like for me to look at? Thanks!

PR #659, opened 2022-08-03
Fix Charset and closing in EvalTimeNorm

kwalcock:
This is getting in the way of testing in other branches, so I'm fixing it here and then will merge it back.


kwalcock:
Copy and paste error

PR #661, opened 2022-08-03
Update for Scala 2.13 compatibility

kwalcock:


kwalcock:
This is undergoing additional testing.  See also #658.

kwalcock:
This is compiling without any deprecation warnings.  That has been achieved in part by adding ~134 instances of .toIndexedSeq based on the deprecation warning
```
Implicit conversions from Array to immutable.IndexedSeq are implemented
by copying; Use the more efficient non-copying ArraySeq.unsafeWrapArray or an
explicit toIndexedSeq call [to the inefficient method that copies].
```
That last part that I added seems pretty important.  As far as I can tell, these arrays are copied in Scala 2.13 via

```scala
new ArrayOps(xs).toIndexedSeq
immutable.ArraySeq.unsafeWrapArray(Array.copyOf(xs, xs.length))
```
Furthermore, the explicit .toIndexedSeq will cause a copy operation in Scala 2.12 in which the Array is actually turned into a Vector.  Previously we could get away with the Scala 2.12 implicit conversion which turned the Array into a WrappedArray, which does not involve a copy.

The other option, ArraySeq.unsafeWrapArray is in Scala 2.13 either
```
scala.collection.immutable.ArraySeq.unsafeWrapArray
scala.collection.mutable.ArraySeq.unsafeWrapArray
```
while in Scala 2.12 it is
```
import scala.collection.compat.immutable.ArraySeq.unsafeWrapArray
import scala.collection.mutable.ArraySeq.unsafeWrapArray
```
and either way they are both too long.  I think that issues like these were keeping this branch from being merged previously and it may have been waiting for more clarity based on requirements of Scala3 and how arrays are passed to DyNet.  At just this minute I am preferring an implicit conversion via a .toWrapped method rather than .toIndexedSeq.  That might give us more flexibility and control.  I'll do some experiments and post examples.

kwalcock:
FYI @thebrenthaines.  If you know otherwise or have a different solution, please chime in.

kwalcock:
Since we have to use something like toIndexedSeq because of the deprecation but not that because it will unnecessarily make inefficient copies, I suggest the design below.  In order to add an operation to Array, it is unfortunately necessary to use an import. so code would change from before Scala 2.13
```scala
   val tags = entity.document.sentences(entity.sentence).tags
```
to inefficient Scala 2.12 and 2.13
```scala
   val tags = entity.document.sentences(entity.sentence).tags.get.toIndexedSeq
``` 
to efficient (but slightly unsafe) Scala 2.13
```scala
import org.clulab.scala.WrappedArray
   val tags = entity.document.sentences(entity.sentence).tags.get.toWrapped
```
The code to do this is for Scala 2.11 & 2.12
```scala
package object scala {

  implicit class WrappedArray[T: _root_.scala.reflect.ClassTag](array: Array[T]) {

    def toWrapped: _root_.scala.collection.IndexedSeq[T] = Predef.genericWrapArray(array)
  }
}
```

and for Scala 2.13
```scala
package object scala {

  implicit class WrappedArray[T: _root_.scala.reflect.ClassTag](array: Array[T]) {

    def toWrapped: _root_.scala.collection.immutable.IndexedSeq[T] =
        _root_.scala.collection.immutable.ArraySeq.unsafeWrapArray(array)
  }
}
```
The same procedure seems to work for Scala 3 as well.

That means I have to go back and replace all those .toIndexedSeqs that were just added with .toWrapped.

In the longer term we should probably be replacing Array with ArraySeq where possible.  This may cause hassles with  Java/Python interop and serialization and it necessitates a dependency on the compatibility library even for simple programs, etc.  Perhaps we can save it for projects that target Scala 3 directly and not use it for Scalal 2 projects that get promoted to 3.

kwalcock:
I think there may be a better option and am working on it.

MihaiSurdeanu:
Let me know when I should look at this.

kwalcock:
There seems to be a way to provide an overriding, implicit conversion on a per-file basis so that all the .toIndexedSeq is unnecessary.  One import just needs to be added to the top of each file that should be optimized away from the default conversion.  Here's an example App
```scala
package com.keithalcock.unsafeWrapArray

import com.keithalcock.unsafeWrapArray.WrappedArray._

object UnsafeWrapArrayApp extends App {
  val array = Array(1, 2, 3, 4, 5)

  def process(indexedSeq: IndexedSeq[Int]): Unit = println(indexedSeq)

  process(array)
}
```
and the utility that makes it work for Scala 2.13, 3.0, and 3.1.
```scala
package com.keithalcock.unsafeWrapArray

import scala.collection.immutable.ArraySeq
import scala.language.implicitConversions

object WrappedArray {

  // Undo the standard conversion to prevent ambiguity from LowPriorityImplicits2 resulting in error
  // "Note that implicit conversions are not applicable because they are ambiguous:".
  // See https://stackoverflow.com/questions/15592324/how-can-an-implicit-be-unimported-from-the-scala-repl.
  def copyArrayToImmutableIndexedSeq[T](xs: Array[T]) = ()

  implicit def toIndexedSeq[T](array: Array[T]): IndexedSeq[T] = {
    println("This is my implicit conversion!")
    ArraySeq.unsafeWrapArray(array)
  }
}
```
The one for Scala 2.11 and 2.12 looks similar.  I'll probably wait a couple of days to make sure nobody on Scala support groups takes issue with it, because despite being just one line it's a lot of changes to undo if wrong.

MihaiSurdeanu:
> There seems to be a way to provide an overriding, implicit conversion on a per-file basis so that all the .toIndexedSeq is unnecessary. One import just needs to be added to the top of each file that should be optimized away from the default conversion. Here's an example App
> 
> ```scala
> package com.keithalcock.unsafeWrapArray
> 
> import com.keithalcock.unsafeWrapArray.WrappedArray._
> 
> object UnsafeWrapArrayApp extends App {
>   val array = Array(1, 2, 3, 4, 5)
> 
>   def process(indexedSeq: IndexedSeq[Int]): Unit = println(indexedSeq)
> 
>   process(array)
> }
> ```
> 
> and the utility that makes it work for Scala 2.13, 3.0, and 3.1.
> 
> ```scala
> package com.keithalcock.unsafeWrapArray
> 
> import scala.collection.immutable.ArraySeq
> import scala.language.implicitConversions
> 
> object WrappedArray {
> 
>   // Undo the standard conversion to prevent ambiguity from LowPriorityImplicits2 resulting in error
>   // "Note that implicit conversions are not applicable because they are ambiguous:".
>   // See https://stackoverflow.com/questions/15592324/how-can-an-implicit-be-unimported-from-the-scala-repl.
>   def copyArrayToImmutableIndexedSeq[T](xs: Array[T]) = ()
> 
>   implicit def toIndexedSeq[T](array: Array[T]): IndexedSeq[T] = {
>     println("This is my implicit conversion!")
>     ArraySeq.unsafeWrapArray(array)
>   }
> }
> ```
> 
> The one for Scala 2.11 and 2.12 looks similar. I'll probably wait a couple of days to make sure nobody on Scala support groups takes issue with it, because despite being just one line it's a lot of changes to undo if wrong.

Very nice @kwalcock !

thebrenthaines:
> There seems to be a way to provide an overriding, implicit conversion on a per-file basis so that all the .toIndexedSeq is unnecessary. One import just needs to be added to the top of each file that should be optimized away from the default conversion.

This is a great application of implicits. I know of no better approach for achieving your goals, @kwalcock.

kwalcock:
This PR is being replaced by another.

PR #662, opened 2022-08-09
Make sure BIO values are legit

kwalcock:
This fixed the problem in the test case and in the 100 habitus files


kwalcock:
See also https://github.com/clulab/habitus/issues/161.

MihaiSurdeanu:
LGTM @kwalcock !

Issue #663, opened 2022-08-10
NumericEntityRecognition hangs


kwalcock:
This is being transferred from https://github.com/clulab/habitus/issues/160.  This particular sentence was taken from one of the problematic files and added to a unit test:

Table 2 : Development forecasts for the dry cold season 2020_2021 speculations Dagana Lake Podor Matam Bakel Valley Tomato Onion but Yam Okra others 1,016 960 250 80 150 1,300 50 1,200 50 700 2,300 3,800 225 800 2 150 200 3,691 7,160 3 130 1,480 1,136 2,515 300 130 875 1,050 30 365400 3,775 Total 2487 3,450 8,575 825 19 112 Source : LM 12 / Saed The SSF 2020/2021 campaign is timidly being set up on the entire left bank of the Senegal River with the establishment of nurseries .

kwalcock:
The main problem seems to be this rule in `measurements.yml` that is prohibitively expensive to run on long lists of numbers.  The sentence in the second file that "hangs" is similar to the previous one.  We're probably going to need to worry about the computational complexity of our rules.

Table 2 : Development forecasts for the dry cold season 2020_2021 speculations Dagana Lake Podor Matam Bakel Valley Tomato Onion but Yam Okra others 1,016 960 250 80 150 1,300 50 1,200 50 700 2,300 3,800 225 800 2 150 200 3,691 7,160 3 130 1,480 1,136 2,515 300 130 875 1,050 30 365400 Total 2487 3,450 8,575 3,775 825 19 112 Source : LM 12 / Saed in Matam , NPK and urea needs are estimated , harmonized with the Apov committee , then sent to the DRDR since the end of August .

```
  - name: measurement-3
    label: Measurement
    priority: ${ rulepriority }
    type: token
    action: mkSharedMeasurementMention
    pattern: |
      (@number:Number [!tag = /^NN|LRB/]{0,2} [word = /,|and|to/]*)+ @number:Number @unit:MeasurementUnit
 ```

**Comment by MihaiSurdeanu** [comment]
Thanks @kwalcock !
How long does it take to process this sentence with and without the rule above?

**Comment by kwalcock** [comment]
It has never finished, so I don't know, but probably at least 15 minutes.  In the meantime I have re-written the rule to
```
      (@number:Number ([tag = IN] [tag = CD])? [!tag = /^NN|LRB|CD/]{0,2} [word = /,|and|to/]*)+ @number:Number @unit:MeasurementUnit
```
[...]

PR #664, opened 2022-08-10
Adjust measurement-3 rule

kwalcock:
...so that it can be applied in a reasonable amount of time with results that pass tests, particularly on long lists of numbers.

### DIFF
--- a/main/src/main/resources/org/clulab/numeric/measurements.yml
+++ b/main/src/main/resources/org/clulab/numeric/measurements.yml
@@ -22,7 +22,7 @@ rules:
     type: token
     action: mkSharedMeasurementMention
     pattern: |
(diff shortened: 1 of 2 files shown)


kwalcock:
See also #663.

kwalcock:
@maxaalexeeva, it looks like you worked on this rule and these tests, so you might want to weigh in on the change or at least be aware of it.  I did the quickest, easiest thing I could think of, so there is undoubtedly room for improvement.  See #663 for more info.  Thanks!

maxaalexeeva:
Thanks @kwalcock that's good to know and I like the rule this way! I have to admit, I have not thought about test efficiency at all until now!

maxaalexeeva (approved):

PR #666, opened 2022-08-30
not allowing year-like values to share units with other values

maxaalexeeva:
Not allowing year-like values to share units with other values. Example of values sharing units:
```
yields were set to 7.9 and 7.1 t/ha
```
We want to avoid 1999 being associated with the unit `t ha-1`:
```
average yield reached 72 t ha-1 in 1999 and 82 t ha-1 in 2000
```

### DIFF
--- a/main/src/main/resources/org/clulab/numeric/measurements.yml
+++ b/main/src/main/resources/org/clulab/numeric/measurements.yml
@@ -22,7 +22,7 @@ rules:
     type: token
     action: mkSharedMeasurementMention
     pattern: |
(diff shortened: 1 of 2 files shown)


MihaiSurdeanu:
This makes sense. But is there a good reason for allowing `[tag = IN]` in the rule?

kwalcock (approved):
Code looks fine.  Rule looks complicated, probably by necessity.

kwalcock:
I *think* that Mihai OK'd this at the meeting yesterday, but I'll wait just a little longer in case someone contradicts.

MihaiSurdeanu:
Ok to merge!

PR #667, opened 2022-08-30
Overlap NamedEntities more carefully

kwalcock:


kwalcock:
@maxaalexeeva, would you like to try this out to see if it fixes the problems you were seeing with the locally published master branch?  This lets longer custom named entities overwrite shorter ones.

MihaiSurdeanu:
Thanks @kwalcock !

kwalcock:
I tried this out on Habitus and it did fix the problems.  Habitus with processor master failed 11 tests and after this fix they all went away except for the one that specifically tests for the incorrect behavior.  I'm ready to merge unless @maxaalexeeva has objections.  It's probably a good time to republish as well so we don't get used to the bug in the currently published version.

maxaalexeeva:
@kwalcock thank you! Sorry didn't get to this yet - had to get a couple other things out the way. If tests in main now pass, it's most likely good, but I can check in a couple of hours.

kwalcock:
@maxaalexeeva, waiting is easy enough :-)  Want to be sure it's working the way you expect it to.  It's TestHabitusProcessor behavior of "acronym" that will fail because it is checking for the incorrect behavior.

maxaalexeeva:
@kwalcock I have not found any additional issues! Thank you!

maxaalexeeva:
@kwalcock let me know when you release processors and I'll adjust habitus main (in addition to changes to autocloser and lazy ner, there are a bunch of tests that are supposed to pass with the new processors and I will want to go through them, see if they actually pass)

kwalcock:
@maxaalexeeva, will do.  There are updates to *.md files which means it will take a couple of hours still.

PR #669, opened 2022-09-07
Update for Scala 2.13 with implicit WrappedArray

kwalcock:


kwalcock:
This has compiled, tested, and published without error on Windows, Linux, and Intel and M1 Macs.  I will try to use it with a few projects that depend on it to see if they still run.

kwalcock:
@thebrenthaines, I'd be interested to know whether this works for you.

thebrenthaines:
> @thebrenthaines, I'd be interested to know whether this works for you.

Hi @kwalcock, I'll give it a try today and let you know. Thanks!

kwalcock:
@thebrenthaines, any luck?  I almost have a version for Scala 3 working.

thebrenthaines:
> @thebrenthaines, any luck? I almost have a version for Scala 3 working.

Hi @kwalcock, sorry about that. I was wrapped up in other fun work (wrapping javacv in a reasonably well-typed scala package) and neglected to test in a timely way. Your gentle nudge did the trick though. I was able build and incorporate your branch seamlessly. No changes. It was sort of anticlimactic, really. I had to triple check that I was pulling in the right library.

Anyway, it looks good to me. Thanks for putting it together!

kwalcock:
@MihaiSurdeanu, this is seeming to work and it's up to date with the current master branch.  There are lots and lots of changes, but they are mostly touching up old syntax.  Notable are the Scala version specific directories that you might want to be aware of.  The ThreadUtils was a pre-existing issue.  The DependencyMap and GraphMap needed to be pulled out of scala to account for changes in Scala 2.13, but are otherwise little changed.  package.scala is used to rename some types.  The Wrapper classes are empty for Scala 2.11 and 2.12 which means that functionality should not be affected at all.  In the Scala 2.13 version, some implicit trickery is used to override the copy that would otherwise be performed whenever an Array is converted into a Seq.  The Arrays get wrapped similarly to how it is done in 2.12 and 2.11: somewhat unsafely.

Presently, all three Scala versions (2.11, 2.12, and 2.13) will be tested by Jenkins.  If this turns out to be a problem because of the wait or because people find it difficult to write syntax compatible with all three versions, the the 2.13 can be removed and I will make any needed touch ups before publication.

I think this is ready to go.

MihaiSurdeanu:
Yay! Thank you @kwalcock !
Ok to merge, as far as I'm concerned.

PR #673, opened 2022-11-08
Address Odin bug related to Inst equality and hashing

kwalcock:


MihaiSurdeanu:
Thanks @kwalcock ! LGTM!

PR #674, opened 2022-11-09
Initialize DyNet automatically

kwalcock:
It's a single line in object CluProcessor.  So far it seams like a reasonable place to put it.


MihaiSurdeanu:
Nice! Ready to merge!

myedibleenso:
If you need to initialize DyNet with additional memory for whatever reason prior to instantiating a `CluProcessor`, this change will respect your memory settings, right?

kwalcock:
The first initialization wins, as was the case before, so just make sure to be first in line.  That includes things potentially budging by referring to the `object CluProcessor` in round-about ways.  There is a chance of mysterious behavior.

PR #675, opened 2022-11-13
Added support for keeping track of "section names"

enoriega:
Added a new optional array field to the sentence object called `sections`. This optional field is meant to keep a hierarchy of section names to which the sentence belongs in its corresponding document.

`sections` values are assigned the by client code, not a processor instance, and its interpretation is left to the user of the field. The main use case for this is REACH, where we can keep track of the section names from the XML structure.

This PR adds support also for efficient serialization using `DocumentSerializer` as well as a test case. 

The code is backwards compatible and no existing code base should break if this PR is merged.

### DIFF
--- a/main/src/main/scala/org/clulab/processors/Sentence.scala
+++ b/main/src/main/scala/org/clulab/processors/Sentence.scala
@@ -43,6 +43,11 @@ class Sentence(
   var graphs: GraphMap = new GraphMap
   /** Relation triples from OpenIE */
   var relations:Option[Array[RelationTriple]] = None
+  /**
+    * Hierarchy of section names to which this sentence belongs.
+    * The first element is inner-most sub-section name, and the last is the top-most section name
+    */
+  var section:Option[Array[String]] = None
 
 
   def size:Int = raw.length
--- a/main/src/main/scala/org/clulab/processors/Sentence.scala
+++ b/main/src/main/scala/org/clulab/processors/Sentence.scala
@@ -47,7 +47,7 @@ class Sentence(
     * Hierarchy of section names to which this sentence belongs.
     * The first element is inner-most sub-section name, and the last is the top-most section name
     */
-  var section:Option[Array[String]] = None
+  var sections:Option[Array[String]] = None
 
 
   def size:Int = raw.length
@@ -97,7 +97,8 @@ class Sentence(
     val h7 = mix(h6, getAnnotationsHash(norms))
     val h8 = mix(h7, getAnnotationsHash(chunks))
     val h9 = mix(h8, if (dependencies.nonEmpty) dependencies.get.equivalenceHash else None.hashCode)
-    finalizeHash(h9, 10) 
+    val h10 = mix(h9, getAnnotationsHash(sections))
+    finalizeHash(h10, 11)
   }
 
   /**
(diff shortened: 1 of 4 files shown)


main/src/main/scala/org/clulab/processors/Sentence.scala
```diff
@@ -190,18 +196,19 @@ object Sentence {
...
+             words: Array[String],
+             tags: Option[Array[String]],
+             lemmas: Option[Array[String]],
+             entities: Option[Array[String]],
+             norms: Option[Array[String]],
+             chunks: Option[Array[String]],
+             tree: Option[Tree],
+             deps: GraphMap,
+             relations: Option[Array[RelationTriple]],
+             sections: Option[Array[String]]
```
kwalcock, line 211: I can image that someone would like a default value ` = None` here so that client code won't need to change.  That's not me, but for the record...

main/src/main/scala/org/clulab/serialization/DocumentSerializer.scala
```diff
@@ -366,9 +370,25 @@ class DocumentSerializer extends LazyLogging {
       os.println(START_RELATIONS + SEP + relations.length)
       relations foreach (t => saveRelationTriple(t, os))
     }
+    // Added by Enrique to support sections names
```
kwalcock, line 373: Do we have to be concerned about the `JSONSerializer` not knowing about the sections?  Perhaps at least a comment could be placed with the `sections` field to alert people that only the `DocumentSerializer` is aware of it so far.

kwalcock (commented):
I wish that a `Document` or processors in general had a better way to deal with subclassed `Sentences` and `Mentions`, but it doesn't yet.  processors is older than I am and there are dependent projects that I don't even know about, so Mihai should look carefully, obviously.  Thanks, Enrique!

main/src/main/scala/org/clulab/processors/Sentence.scala
enoriega: This one is easy. Thanks!

main/src/main/scala/org/clulab/serialization/DocumentSerializer.scala
enoriega: @kwalcock I  see there are two Json serialization classes:

- `org.clulab.serialization.json.JSONSerializer`
-` org.clulab.serialization.json4StopGraph.JSONSerializer`

Are both of them still relevant? 
I can't find the deserialization part for neither of them. I suspect they're abandoned in favor of the `DocumentSerializer`

kwalcock: IDK.  The output could be used by a separate project that we don't know about.

MihaiSurdeanu: I think both serializations need to be supported for now...

MihaiSurdeanu:
This LGTM!

@kwalcock: we can discuss reorganizing the Sentence and Document classes. However, this simple design is on purpose, so students access this information easily without worrying about fancy data structures.

main/src/main/scala/org/clulab/serialization/DocumentSerializer.scala
enoriega: @kwalcock thanks!
@MihaiSurdeanu do you know if both json serializers are relevant today? And if not, which one is it?

MihaiSurdeanu: I know @myedibleenso implemented both (or one?) of them, and he was using them. @myedibleenso: do you have an answer to @enoriega's question?

enoriega:
FWIW, I agree with this observation too. It would be nice to have custom, domain specific fields that are maintained by the developer of the client code (I.e. whoever extends Reach and needs nxml specific data in the doc)

kwalcock:
@enoriega, do you have code that makes use of the new field?  Perhaps we could figure out a way to add the new field to CustomSentence and with something like an implicit conversion from regular Sentence to CustomSentence, make it easy to use in a way that others can imitate.

enoriega:
@kwalcock Yes. Reach's #775  is the sister PR of this one. You can specifically look at: 
https://github.com/clulab/reach/pull/775/commits/a31d349aaadcbee9c25684e04dee984e13c0f8f3

I think the only challenge will be to make the serialization code aware of the custom fields. This is used when pre-annotating documents with REACH.

kwalcock:
It's a bit tentative, but I *think* this functionality is being handled by changes local to reach.  The PR can be reopened if necessary.

PR #678, opened 2022-11-18
Make use of control strings optional

kwalcock:
Something this simple might suffice.


MihaiSurdeanu:
LGTM. Thanks @kwalcock !

PR #685, opened 2022-11-30
Split a text into smaller pieces and make a single document of them.

kwalcock:
This is a draft because the changes were concentrated in Processor when they really should be distributed between Document and Sentence.  However, it shows that the processors project does not necessarily need to be changed at all to achieve the functionality.  The test cases should show the intended use.  This is related to #676, but doesn't necessarily replace it.  I should double check that no offset values have been missed.


MihaiSurdeanu:
LGTM @kwalcock. Thanks!

kwalcock:
@myedibleenso, [this](https://github.com/clulab/processors/blob/kwalcock/multidocument/main/src/test/scala/org/clulab/processors/TestMkCombinedDocument.scala) shows a couple of additional ways that your `<br>` might be handled.  Perhaps you can say whether it seems useful or not.

kwalcock:
This shouldn't change any behavior or affect performance, so I'd like to add it if only to preserve the idea, whether or not @myedibleenso or anyone ever uses it.  I'm less certain about the alternative, #676.  Perhaps it can be looked at again now that it has been mulled over for a couple of months.

PR #687, opened 2022-12-03
Model UnitNormalizer on SeasonNormalizer

kwalcock:


kwalcock:
@maxaalexeeva, I *think* this is what was described in recent emails.  I'm not sure to what extent you want to change the file or turn the functionality on or off.  That's isn't especially accounted for yet, but can probably be added.  I did change your one dictionary per unit to just a case class, which should be much smaller, and held off with UNK until the very end at which time it wasn't necessary anyway.  Feel free to change to your liking or let me know what to do.

MihaiSurdeanu:
Looks great @kwalcock !

maxaalexeeva:
Thanks @kwalcock ! I'll have a look tomorrow!

PR #689, opened 2022-12-05
Fix HolidayNormalizer

kwalcock:


maxaalexeeva:
Neat! Thanks! I'll check tomorrow!

maxaalexeeva:
@kwalcock turns out the error was because ", " is an unfortunate separator for the `// unit, class` comment in the measurement-unit.tsv file---`Jr. Day` in `//	Martin Luther King, Jr. Day` (holidays.tsv) is not a class...

Gonna use a different separator.

PR #691, opened 2022-12-12
adding unit classes

maxaalexeeva:
unit entities are returned with a more specific class label, e.g., MEASUREMENT-`AREA`, MEASUREMENT-`DURATION`

Note: some units don't have a class; in that case, they are just `MEASUREMENT`


maxaalexeeva:
@kwalcock the tests don't seem to be running?

kwalcock:
I saw that.  Maybe they were just slow to start.

kwalcock (approved):
LGTM.  Shortest review yet!

kwalcock:
This may be needed for a SEC update.  The resulting master would need to be published, either to maven or artifactory.

kwalcock:
Does @MihaiSurdeanu think it's OK?

MihaiSurdeanu:
LGTM too!

PR #696, opened 2023-01-10
Smooth out greedyPredict and inferenceWithScores

kwalcock:


kwalcock:
I must have grepped one day for ArrayBuffer and was compelled to do some housekeeping.  *If* I didn't make any mistakes, then no behavior should have changed.

kwalcock:
If something goes wrong, this will have been an unforced error.  On the other hand, it's not good to be afraid to change the code.  I'll go ahead and request review.

PR #700, opened 2023-01-13
Get versioning, especially of library dependencies, sorted out

kwalcock:
This should be compatible with the habitus and sec projects.  It is syntactically compatible with habitus, but tests fail.

### DIFF
--- a/main/src/main/scala/org/clulab/numeric/UnitNormalizer.scala
+++ b/main/src/main/scala/org/clulab/numeric/UnitNormalizer.scala
@@ -9,10 +9,20 @@ import scala.io.Source
 
 case class NormAndUnitClass(norm: String, unitClassOpt: Option[String])
 
-class UnitNormalizer(unitNormalizerPath: String) {
-  val normMapper: Map[String, NormAndUnitClass] = UnitNormalizer.readNormsFromResource(unitNormalizerPath)
-
+abstract class BaseUnitNormalizer() {
   def mkUnit(text: Seq[String]): String = text.mkString(" ").toLowerCase()
+  def norm(text: Seq[String]): String
+  def unitClassOpt(text: Seq[String]): Option[String]
+}
+
+object NullUnitNormalizer extends BaseUnitNormalizer() {
+  def norm(text: Seq[String]): String = mkUnit(text)
+
+  def unitClassOpt(text: Seq[String]): Option[String] = None
+}
+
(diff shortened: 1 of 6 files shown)


main/src/main/scala/org/clulab/numeric/UnitNormalizer.scala
```diff
@@ -9,10 +9,20 @@ import scala.io.Source
 
 case class NormAndUnitClass(norm: String, unitClassOpt: Option[String])
 
-class UnitNormalizer(unitNormalizerPath: String) {
-  val normMapper: Map[String, NormAndUnitClass] = UnitNormalizer.readNormsFromResource(unitNormalizerPath)
-
+abstract class BaseUnitNormalizer() {
```
kwalcock, line 12: @maxaalexeeva, here's what I was trying to explain.

main/src/main/scala/org/clulab/numeric/mentions/MeasurementMention.scala
```diff
@@ -15,7 +15,7 @@ class MeasurementMention ( labels: Seq[String],
                            val value: Option[Seq[String]],
                            val unit: Option[Seq[String]],
                            val fromRange: Boolean,
-                           unitNormalizer: UnitNormalizer)
+                           unitNormalizer: BaseUnitNormalizer = NullUnitNormalizer)
```
kwalcock, line 18: @maxaalexeeva, there is a default value so that habitus is at least syntactically compatible.

kwalcock:
@maxaalexeeva, if I connect this version of processors to habitus, quite a few test cases fail.  Habitus main is usually connected to processors 8.5.2.  Would you expect changes (other than the ones I made) in 8.5.3 to cause problems?  So far the update is good with SEC, reach, and SKEMA.

maxaalexeeva:
Hi @kwalcock, I am publishing processors locally, and there's some serious download going on, so I'll probably have to look at the failed tests tomorrow. My best guess is that they fail because I did not adjust habitus for the new measurements: a lot of measurement-related rules are in terms of exact strings instead of flexible regexes (e.g., entity='B-MEASUREMENT'), but many of those measurements are not there anymore because they have been modified by unit class (e.g. MEASUREMENT-AREA). Should be easy to fix if that's the case. 

Sorry, I am not sure how I missed that that change needs to happen.

maxaalexeeva:
That was indeed the case. See https://github.com/clulab/habitus/pull/176.

PR #707, opened 2023-01-25
Prepare OdinStarter for template

kwalcock:


kwalcock:
It would be handy to keep a copy of `OdinStarter` in the processors project if it doesn't clutter things up too much.  If these files are only included in the starter template repo, they are obscured with one level of templating redirection.  All the resource files have to be escaped properly and there is not a good way to track changes required by updates to processors.  If the file is here, then I can note when it needs to change and propagate the changes to the template.

Liberties with the original file and they can definitely be undone. 

* It seems like a good idea to follow GitHub's lead and change from `master` to `main`.  
* As was noted in a comment, the parallel sequences of kbs and the case information was not obvious.  I changed the data structure to force pairs and hope it is easier to understand, but it may not be.
* It didn't seem like the defs were necessary, so they were changed to vals.
* The printing of the mentions was made a little prettier so that the output can be more easily interpreted.

For people new to Scala, it might be useful to show a Scala 3 version with the Python formatting.

MihaiSurdeanu:
I agree with all these points. Thanks @kwalcock !

kwalcock:
The first test run failed at `it should "collapse prepositions with verb objects"` in `org.clulab.processors.TestEnhancedSemanticRoles`, even though nothing added here would have affected a test.  The second run passed without any changes having been made.  This may be the case of an "unreliable" test and is being filed as an issue.

kwalcock:
The `OdinStarter` code here was added to the [sbt-processors-small.g8](https://github.com/clulab/sbt-processors-small.g8) repo and used as a template for projects on Windows, Linux, and Mac computers which ran all the tests and sample programs on Scala 2.11, 2.12, 2.13, and 3.1 (excepting the one using the discontinued `args`).  On Linux it worked with Java 1.8 and 11 and on Mac it worked with Java 1.8 Intel, 1.8 M1, and 18 Intel.  Not that it has never crashed before...

PR #709, opened 2023-01-31
Add a webapp

kwalcock:
This is still being fiddled with in order to make it reusable from other projects.  It is presently already useful in this project.


kwalcock:
The webapp works now from webapp/run, from the docker container generated by dockerizeWebapp, and as a libraryDependency generated by publish or publishLocal.

kwalcock:
See also the readme at  https://github.com/clulab/processors/tree/kwalcock/webapp/webapp.

MihaiSurdeanu:
This is fantastic @kwalcock !
Very very useful.
My only comment is: is all the brat subdir needed? For example, it has a folder called "casestudies", which has a lot of content that doesn't seem necessary?

kwalcock:
Good point!  I copied it over from Eidos without second thought.  It might have been superfluous even there.  I will check.

PR #715, opened 2023-02-17
Change from AutoCloser to Using

kwalcock:


kwalcock:
Neither Scala 2.11 or 2.12 had Using, so the AutoCloser was a local (and now temporary) substitute.  For the Scala 2.13 and 3 version of processors, the scala-collection-compat library was needed.  Despite Using being unrelated to collections, the library does include a backport of Using.  Since that's what will be used in the future, I'd rather not have people getting used to AutoCloser or copying it around to other code when it is no longer necessary.  The class remains in case it is used in dependent projects, but processors no longer needs it itself.  For Scala 2.11 there still needs to be a hack to be able to close a Source.  This is why the import is org.clulab.scala.Using._ instead of scala.utils.Using.  If or when the compat library is updated to include Source closing, the import statements can be changed.  I filed an issue with them.  There's also something similar in fatdynet that needs to be updated.

MihaiSurdeanu:
This looks nice!
One request: can you please add a documentation page (the project's GH pages?) with the recommended way to access resources in processors (and our lab) based on this?
Thanks!

kwalcock:
Documentation has been added at https://github.com/clulab/processors/wiki/Releasing-resources-(e.g.,-closing-files)

MihaiSurdeanu (approved):
Nice! This was a lot of work... Thanks!

PR #716, opened 2023-02-22
Show how letters and words can be veiled

kwalcock:


kwalcock:
There were tiny changes made to some existing processors classes, but only for convenience, and the webapp output was beautified.  Otherwise, the functionality lies in Veil.scala, which can be scrapped or sit there harmlessly if need be.

MihaiSurdeanu:
This feels very complicated. I think I'd like an example before I'm convinced...

kwalcock:
There are two examples ([veilText](https://github.com/clulab/processors/blob/1d22a7758916fb01c082c8734dc85e50b93714db/main/src/main/scala/org/clulab/processors/clu/Veil.scala#L205-L218), [veilDocument](https://github.com/clulab/processors/blob/1d22a7758916fb01c082c8734dc85e50b93714db/main/src/main/scala/org/clulab/processors/clu/Veil.scala#L220-L234)) at the bottom of [Veil.scala](https://github.com/clulab/processors/blob/kwalcock/veil/main/src/main/scala/org/clulab/processors/clu/Veil.scala).  Would those suffice?

MihaiSurdeanu (approved):
Nice!

Just to be safe, we should ask @myedibleenso for his feedback as well.

kwalcock:
This PR has been updated with the changes to master.  In review, it includes

* small refactorings in Document and Sentence that allow internals to be manipulated more easily
* the veiling functionality in an independent class
* two examples of veiling
* some cosmetic webapp improvements (which I will probably move to a different PR)

@myedibleenso, I'd like to get this merged as soon as you can look at it.

Unless something is wrong with the changes to Document and Sentence, this code will probably just sit around until someone needs it.  In a PR is not a good place to sit, though.

main/src/main/scala/org/clulab/processors/clu/Veil.scala
```diff
@@ -0,0 +1,244 @@
...
+
+  /** Treat this text as if the words "( Hahn-Powell , 2012 )" did not exist
+    * for the purpose of annotate, but do include them in the document.
+    */
+  def veilDocument(processor: Processor): Unit = {
+    val text = "To be loved by unicorns is the greatest gift of all (Hahn-Powell, 2012)."
+    val document = processor.mkDocument(text)
+    val veiledWords = Seq((0, Range.inclusive(document.sentences(0).raw.indexOf("("), document.sentences(0).raw.indexOf(")"))))
+    val veiledDocument = new VeiledDocument(document, veiledWords)
+    val annotatedDocument = veiledDocument.annotate(processor)
```
myedibleenso, line 231: Would be good to add a type hint here to make it clear that `annotatedDocument` is a `Document`.

myedibleenso (approved):
Yikes!  Sorry, @kwalcock .  I somehow lost track of this PR from ages ago.  Thank you for the reminder.  I think this is going to be very useful for a bunch of tasks.

kwalcock:
Good point.  It's even better than a hint: it's enforced.

MihaiSurdeanu:
I think this is ready to be merged. Thanks @kwalcock !

Issue #309, opened 2019-02-18
unable to dynamically load a new rule requiring resources or variables


michaelcapizzi:
I'd like to be able to dynamically add a new rule to the `ExtractorEngine`.  

I know how to do this with `readSimpleFile()`, but if the rule in question requires resources (e.g. `Word2Vec`) or variables, then it must use `readMasterFile()`.  But this *rereads* all resources and variables each time it is called.  And if the `Word2Vec` file is large, this can be very slow.

I tried to `extend` `RuleReader` and then `override` `readMasterFile()` so that it can optionally take an existing `OdinConfig`:

```scala
/**
  * Custom Rule Reader that *stores* the [[OdinConfig]]
  */
class CustomRuleReader(
                      actions: Actions,
                      charset: Charset
                      ) extends RuleReader(actions, charset) {

  // empty config
  var config = OdinConfig(resources = OdinResourceManager(Map.empty))

  /**
    * non-private form of readResources
    */
  def readResources(data: Map[String, Any]): OdinResourceManager = {
    //println(s"resources: ${data.get("resources")}")
    val resourcesMap: Map[String, String] = data.get("resources") match {
      case Some(m: JMap[_, _]) => m.asScala.map(pair => (pair._1.toString, pair._2.toString)).toMap
      case _ => Map.empty
    }
    OdinResourceManager(resourcesMap)
  }

  /**
    * non-private form of readTaxonomy
    */
  def readTaxonomy(data: Any): Taxonomy = data match {
    case t: Collection[_] => Taxonomy(t.asInstanceOf[Collection[Any]])
    case path: String =>
      val url = mkURL(path)
      val source = Source.fromURL(url)
      val input = source.mkString
      source.close()
      val yaml = new Yaml(new Constructor(classOf[Collection[Any]]))
      val data = yaml.load(input).asInstanceOf[Collection[Any]]
      Taxonomy(data)
  }

  /**
    * non-private form of readRules
    */
  def readRules(
                 rules: Collection[JMap[String, Any]],
                 config: OdinConfig
               ): Seq[Rule] = {

    // return Rule objects
    rules.asScala.toSeq.flatMap { r =>
      val m = r.asScala.toMap
      if (m contains "import") {
        // import rules from a file and return them
        importRules(m, config)
      } else {
        // gets a label and returns it and all its hypernyms
        val expand: String => Seq[String] = label => config.taxonomy match {
          case Some(t) => t.hypernymsFor(label)
          case None => Seq(label)
        }
[...]

PR #717, opened 2023-02-22
Add CustomRuleReader

kwalcock:


kwalcock:
This is meant to address #309

kwalcock:
Changes to existing code are only the substitution of protected for private in one class, which are unlikely to affect anything else.

Issue #718, opened 2023-03-01
Preferred size of graphs is not set upon Document deserialization


kwalcock:
This needs to be done because otherwise the preferred size is determined based on the edges.  If one of the roots is higher than any value in the edges, the root will be lost.  I don't know why we are not using the roots directly.  They are instead recalculated.  Perhaps this is to account for root values that might be bogus, either intentionally or not.  (The definition of root has changed over time.)  One of the unit tests in reach is losing a root after serialization and this seems to be the cause.  A PR is already prepared.

https://github.com/clulab/processors/blob/f14ada88eecfc13e5ed6d6f78aa8f2dca9d84f3b/main/src/main/scala/org/clulab/serialization/json/JSONSerializer.scala#L76-L97

kwalcock:
The roots are for output only so that a human can read them.  However, since they are redundant and have been incorrect or just interpreted differently in the past, they are recalculated when the file gets read in.

kwalcock:
In versions about to me merged, the size gets taken from the number of words rather from the edges.

kwalcock:
Addressed with #719

PR #719, opened 2023-03-02
Fix issues with serialization of documents

kwalcock:
Preferred size of graphs must be given and values of roots sorted


kwalcock:
See #718

PR #720, opened 2023-03-04
Refactor JSON serialization for the sake of reach

kwalcock:
There will be a matching PR for reach that should reduce serialization time in some cases from about 1 day to 10 seconds.

In this PR
* the Odin serialization is moved out of package file into a normal file for sake of clarity
* implicit conversions are made explicit for the same reason
* the base class MentionOps is set up to be a template class from which reach can inherit.  Two important methods are included: asMentionOps and documentEquivalenceHash.  The first converts a Mention to something more specific and the second provides ways in which the hash can be cached.
* the preferred size of graphs is used in their construction
* roots are sorted for consistent output

There will eventually be a wiki page that describes the design of these classes and how they should be extended.


kwalcock:
This has been replaced by #722.

PR #723, opened 2023-03-09
Move all hash code to HashCode class

kwalcock:
Let the computer tally the count needed for finalization and transfer intermediate results from one state to the next.  Also show inconsistencies in how hashes are being used.

### DIFF
--- a/main/src/main/scala/org/clulab/odin/Mention.scala
+++ b/main/src/main/scala/org/clulab/odin/Mention.scala
@@ -1,12 +1,13 @@
 package org.clulab.odin
 
-import scala.util.matching.Regex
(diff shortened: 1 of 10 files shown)


kwalcock:
I hadn't looked closely enough.  It is failing for a specific Scala version.

kwalcock:
This consolidates the mathematics of hashing into a centralized location where it can be better observed and reused.  It brought to light several issues in both processors and reach which should eventually be addressed.  This is part of the effort to get reach to serialize faster, but is optional and should be merged after the rejson branch.  There will probably be merge conflicts at that time.

MihaiSurdeanu (approved):
Very clean! Thanks @kwalcock !

PR #727, opened 2023-03-20
Refactor DependencyUtils so that distToRoot can be reused

kwalcock:


kwalcock:
The inner functions distToRoot and countSteps have been pulled out so that they can be reused elsewhere.  There is also a distToRootOpt that makes it more obvious that the root may not be reachable.

maxaalexeeva (approved):
Looks good to me! Thanks!

kwalcock:
This refactoring was added so that Masha doesn't need to do something the hard way next time.

PR #730, opened 2023-07-07
Run the BalaurProcessor from a resource model

kwalcock:


kwalcock:
The tests can't pass because of the locally published dependency.  This is being merged into a side branch and not main/master, so I will go ahead.

MihaiSurdeanu:
Thank you!

PR #732, opened 2023-07-11
Update collection compatibility library

kwalcock:


kwalcock:
The external `scala-collection-compat` library has been updated to take care of closing a Source, so neither the local Closer or Using are necessary.  All that functionality is now in scala.util.Using.

fatdynet has also been updated similarly and is now at version 0.4.5, which is not yet used here.  If there is to be a last version of processors with fatdynet, then I should update the dependency.  It will remove the need for the `implicit object CloseableModelSaverReleaser` and prevent the eviction notice for `scala-collection-compat` mentioned in the build file.

MihaiSurdeanu (approved):
Thanks @kwalcock

PR #733, opened 2023-07-13
Make cosmetic improvements to webapp

kwalcock:
These were noticed during testing of the Veil, but are independent of it, so they are being submitted separately.


kwalcock:
This was part of the Veil, but should really be done separately.

PR #738, opened 2023-08-21
Fix tests

kwalcock:


kwalcock:
@alicekwak, the problem with the distance to root in the failing tests was unrelated to your changes.  They stem from an earlier change in which the previous version tested good and the changes tested good, but the merged version did not.  It's a weird CI/github kind of problem.

I went ahead and suggested some changes to postProcessNumericEntities.  If either condition from the prevWords was satisfied, the contextWords were calculated unnecessarily at quite some expense.  A regular expression within a loop might be recompiled multiple times, so I moved it to a place where it would only happen once.  There's also a little trick in which a set can act as a function that checks for membership.  It's OK to undo the changes if they are unwanted.

I'm not sure whether a three- or one-digit number should be good in the context.  The window value will result in window words to the left, the word at wordIndex and then window - 1 words to the right so that left and right aren't symmetric.  I don't know if that's the intention.

alicekwak:
> @alicekwak, the problem with the distance to root in the failing tests was unrelated to your changes. They stem from an earlier change in which the previous version tested good and the changes tested good, but the merged version did not. It's a weird CI/github kind of problem.
>
> I went ahead and suggested some changes to postProcessNumericEntities. If either condition from the prevWords was satisfied, the contextWords were calculated unnecessarily at quite some expense. A regular expression within a loop might be recompiled multiple times, so I moved it to a place where it would only happen once. There's also a little trick in which a set can act as a function that checks for membership. It's OK to undo the changes if they are unwanted.
> 
> I'm not sure whether a three- or one-digit number should be good in the context. The window value will result in window words to the left, the word at wordIndex and then window - 1 words to the right so that left and right aren't symmetric. I don't know if that's the intention.

Thanks @kwalcock! I'll review this soon and let you know if I have any questions.

alicekwak (approved):
Thanks @kwalcock for the fixes and revisions! I really liked how you changed the code. I think it's much easier to understand the purpose of the filter now. And I agree with your comment on yearPattern (`"[0-9]{1,4}"`). I can change that in my PR so that it's only allowed if the numbers are either two-digit or four-digit!

kwalcock:
It looks like it was approved, so I'll merge.  It would have been OK for @alicekwak to do as well, since it is going into your branch.  Things can be changed again there if necessary.

PR #743, opened 2023-08-30
Allow subclasses to override tokenizer

kwalcock:
The subclass doesn't know what tokenizer the Balaur processor is going to use, so it constructs itself and the superclass, borrows the tokenizer from the superclass, manipulates it, and then overrides the superclass implementation with the modified version.


MihaiSurdeanu (approved):
Thank you @kwalcock !

PR #744, opened 2023-08-30
Fix some Eisner issues, particularly with bad indexes

kwalcock:


kwalcock:
This is my take on it.  All of the complicated logic about the relative head indexes is isolated to a HeadLabelScore class.  Sorry for the long variable names.  I can dial it back some if necessary.  It should work more efficiently and I hope be more likely correct.

main/src/main/scala/org/clulab/processors/clu/CluProcessor.scala
```diff
@@ -759,6 +760,7 @@ class CluProcessor protected (
     if(doc.sentences.length > 0) {
       assert(doc.sentences(0).tags.nonEmpty)
       assert(doc.sentences(0).entities.nonEmpty)
+      assert(doc.sentences(0).lemmas.nonEmpty)
```
kwalcock, line 763: This is a place where a spot check will not work if certain (long) sentences will not be processed.

MihaiSurdeanu (commented):
You inserted a space after every `if` :)

main/src/main/scala/org/clulab/processors/clu/CluProcessor.scala
MihaiSurdeanu: `CluProcessor` will be deleted.

main/src/main/scala/org/clulab/processors/clu/PostProcessor.scala
```diff
@@ -2,26 +2,27 @@ package org.clulab.processors.clu
...
 import org.clulab.processors.Sentence
 
+import java.util.regex.Pattern
+
 object PostProcessor {
   //
   // Patterns for post-processing corrections
   //
-  val VERSUS_PATTERN = """(?i)^vs\.?$""".r
+  val VERSUS_PATTERN = Pattern.compile("""(?i)^vs\.?$""")
```
MihaiSurdeanu, line 11: Just for my own curiosity, why is this better than `.r`?

MihaiSurdeanu (approved):
LGTM!

main/src/main/scala/org/clulab/processors/clu/PostProcessor.scala
kwalcock: It may not be for a Scala regex.  I see that it keeps track internally of its own compiled pattern, which I had forgotten.  On the other hand, findFirstIn(string).nonEmpty is pretty ugly compared to a boolean matches().  If they added that, and they have in Scala 2.13+, I would probably switch back.  Then it would be VERSUS_REGEX instead.  At least now the casual reader can know that the pattern is compiled and the small overhead of the Scala regex is skipped.  Perhaps it's time to re-evaluate my standard procedure.

MihaiSurdeanu: Ok!

PR #745, opened 2023-08-30
Add chunker resource

kwalcock:


kwalcock:
Something like this then.  No external library is needed.

MihaiSurdeanu (approved):
Perfect! Thanks!

PR #748, opened 2023-09-01
Refine example

kwalcock:


kwalcock:
@navalani, here's an updated Debugger and if nobody watching objects, I'll merge it soon.  You might notice the TestDebugger file that shows the debugger working on actual rules, documents, and Odin.  Changes in Extractor and ExtractorEngine show what I hope is a minimally invasive way to incorporate the debugger into the Odin code.  Then the hard part is getting it to debug, of course.  FYI @MihaiSurdeanu.

PR #749, opened 2023-09-02
Use only Dependency and scrap HeadLabelScore

kwalcock:


MihaiSurdeanu (approved):
Cleaner!

kwalcock:
You are too quick.  Previously this information went through five data structures: PredictionScore coming from the neural network, HeadLabelScore (Tuple3), Dependency, HeadLabel (Tuple2), Edge (in preparation for the graph).  This change removes the HeadLabelScore.  The HeadLabelScore and Dependency did essentially the same thing except for fiddling with the values +/- 1.  Now that's mostly internal to the Dependency.

However, I think the HeadLabel might be removed as well, so that may be done shortly.

kwalcock:
If this works, then there is the PredictionScore coming from the neural network, Dependency, and then Edge going to the graphs.

PR #752, opened 2023-09-14
Fixed the "Fall" filter bug

MihaiSurdeanu:


main/src/main/scala/org/clulab/numeric/actions/NumericActions.scala
```diff
@@ -265,7 +265,7 @@ class NumericActions(seasonNormalizer: SeasonNormalizer, unitNormalizer: UnitNor
     val (seasonMentions, otherMentions) = mentions.partition(m => m.foundBy.contains("season"))
     val (springFall, otherSeasons) = seasonMentions.partition(m => m.text.equalsIgnoreCase("spring") || m.text.equalsIgnoreCase("fall"))
     val trueSeasons = springFall.filter { m =>
-      m.tags.head.contains("NN") && {
+      m.tags.get.head.startsWith("NN") && {
```
kwalcock, line 268: Thank you!

This encourages me to continue denoting Options by name (like tagsOpt) and wonder whether something like the [TagSet](https://github.com/clulab/eidos/blob/master/eidoscommon/src/main/scala/org/clulab/wm/eidoscommon/TagSet.scala) class from Eidos should be ported.  We probably shouldn't be using string operations for classifying tags.

main/src/main/scala/org/clulab/numeric/mentions/package.scala
```diff
@@ -327,11 +327,11 @@ package object mentions {
...
-      DateRangeMention(
-        m,
-        TempEvalFormatter.mkDate(seasonNorm.get.startDay, seasonNorm.get.startMonth,yearStart),
-        TempEvalFormatter.mkDate(seasonNorm.get.endDay, seasonNorm.get.endMonth, yearEnd)
-      )
+      val sd = TempEvalFormatter.mkDate(seasonNorm.get.startDay, seasonNorm.get.startMonth, yearStart)
+      val ed = TempEvalFormatter.mkDate(seasonNorm.get.endDay, seasonNorm.get.endMonth, yearEnd)
+      val dm = DateRangeMention(m, sd, ed)
+
+      dm
```
kwalcock, line 334: Somebody is going to sneak in here and give these variables real names ;-)

main/src/main/resources/org/clulab/numeric/atomic.yml
```diff
@@ -24,6 +24,14 @@ rules:
     pattern: |
       [word=/^(1\d\d\d|20\d\d)$/]
 
+  # weak possible years: 1d, 2d, 3d, 4d, 5d, 6d, 7d, 8d, 9d
+  - name: weakyear
+    label: WeakPossibleYear
+    priority: ${ rulepriority }
+    type: token
+    pattern: |
+      [word=/^(1\d|2\d|3\d|4\d|5\d|6\d|7\d|8\d|9\d)$/]
```
kwalcock, line 33: I wonder why this isn't `[1-9]\d`.

kwalcock (approved):

main/src/main/resources/org/clulab/numeric/atomic.yml
MihaiSurdeanu: Yes, it should be. It started as just [1-2] and [8-9] then I incrementally expanded without optimizing. Please adjust!

main/src/main/scala/org/clulab/numeric/mentions/package.scala
MihaiSurdeanu: :)

PR #754, opened 2023-09-16
Replaced DyNet with a transformers/ONNX/Breeze model

MihaiSurdeanu:
This should only be merged after a new processors release with DyNet.
Before merging we need to update the ONNX model(s) and scala-transformers versions in main/build.sbt.


kwalcock:
I will work on the conflicts.

MihaiSurdeanu:
@kwalcock : I think this branch is as good as it gets for now. Before merging, we need to update the scala-transformers version in main/build.sbt.

PR #756, opened 2023-09-20
Merge master and reconcile differences

kwalcock:


kwalcock:
One test is failing: TestHash.  This is because it makes use of `events.yml` which has
```yml
- name: people-eat-food
  priority: ${rulepriority}
  label: Eating
  example: "John eats cake"
  graph: "hybrid"
  pattern: |
    trigger = [lemma=/eat/ & tag=/^V/]
    food:Food = dobj
    person:Person = nsubj
```
and the Balaur processor doesn't seem to support hybrid graphs.  That file is really there for the OdinStarter, so that won't work as intended.  The graph type can probably be changed, but other users of processors will probably run into problems with their rules.

FYI @MihaiSurdeanu

MihaiSurdeanu:
One more idea: to avoid changing external code, maybe CluProcessor should be implemented as an alias for BalaurProcessor?

kwalcock:
See #759 for implementation of the alias idea.

kwalcock:
This is failing in TestNumericEntityRecognition where there is a timed test at the end.  It is taking approximately 5 times longer to complete than it does in master.  I'm not yet sure why.

MihaiSurdeanu:
Hmm... Other than the "fall" changes, that code should not have changed.

kwalcock:
Both versions should have fall in them.  This branch has the changes from master merged.  I'll probably need to time CluProcessor vs. BalaurProcessor again.

kwalcock:
AFAIK this test has never been a problem before.  The time limit used to be 20 seconds.  I swapped the limit for a println and it showed the following results:

| Scala version | Time |
| --- | --- |
| 2.11 | 0:00:01:10.920 |
| 2.12 | 0:00:00:34.028 |
| 2.13 | 0:00:00:08.969 |
| 3.1 | 0:00:00:09.176 |

I'll try to find out where it's coming from.

kwalcock:
I believe this is outdated.  The same changes seem to have found their way into hexatags and then master starting with https://github.com/clulab/processors/commit/be4a0d084e5dc33b097cc97390fa36b4c4721aac.  I intend to close this without a merge.

PR #759, opened 2023-09-20
Use aliases to convert CluProcessor to BalaurProcessor

kwalcock:
Show examples in TestHash.  HomeController still had a CluProcessor and it still works unchanged.  Only the package file needs to be used, so this is just a draft.


MihaiSurdeanu:
Nice!!

PR #760, opened 2023-10-04
Make it compile

kwalcock:


kwalcock:
@navalani, in case you were having trouble compiling, this shows one way to convince it to do so.  It's probably not the right way, so I wouldn't merge it, but check the files changed for hints.  IntelliJ is coloring `copy()` red, but it is just confused.  That's a gotcha.  Final authority on compilation is sbt.  The copy method will be automatically generated for case classes and IntelliJ doesn't realize that.

BTW, the timer looks good.

PR #765, opened 2023-12-10
Added code to generate embeddings

RazvanDu:
The code can generate custom sub-word embeddings based on a model type and a data set.

### DIFF
--- a/main/src/main/python/embedding_generator/main.py
+++ b/main/src/main/python/embedding_generator/main.py
@@ -6,6 +6,7 @@ import argparse
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--model_type', dest='model_type', type=str)
+parser.add_argument('--seed', dest='seed', type=int)
 parser.add_argument('--tokenizer', dest='tokenizer', type=str)
 parser.add_argument('--context_size', dest='context_size', type=int)
 parser.add_argument('--device', dest='device', type=str)
@@ -23,11 +24,18 @@ huggingface_dataset = args.huggingface_dataset # wikipedia
 huggingface_subpart = args.huggingface_subpart # wikipedia
 number_of_tokens = int(args.number_of_tokens) # 10000000
 output = str(args.output) # deberta_embd
+seed = -1 # 42
+
+if args.seed:
+    seed = int(args.seed) # deberta_embd
 
 torch.set_default_device(device)
 
 model = DebertaModel.from_pretrained(model_type).to(device)
 tokenizer = AutoTokenizer.from_pretrained(tokenizer)
+
+if seed != -1:
+    np.random.seed(seed)
 
 if huggingface_dataset:
     if huggingface_subpart:
@@ -38,7 +46,6 @@ else:
     dataset = load_dataset('json', data_files={'train': 'text_data.json'})
 
 tokens_so_far = 0
-crr = 0
 
 sum_tensors = {}
 crr_tensors = {}
@@ -49,7 +56,7 @@ print(len(dataset))
 
 while tokens_so_far < number_of_tokens:  # 10 million
 
-    index = np.random.randint(0, len(dataset))
+    index = np.random.randint(0, len(dataset['train']))
 
     text = dataset['train'][index]['text']
 
@@ -59,16 +66,12 @@ while tokens_so_far < number_of_tokens:  # 10 million
 
     if len(tokens) <= context + 1:
 
-        crr += 1
         tokens_so_far += len(tokens)
-
         sliced_output = {key: value[:, :] for key, value in tokenized.items()}
 
     else:
 
         i = np.random.randint(0, len(tokens) - context - 1)
-
-        crr += 1
         tokens_so_far += context
 
         sliced_output = {key: value[:, i:i+context] for key, value in tokenized.items()}
@@ -101,6 +104,7 @@ for unseen in tokenizer.get_vocab():
     crr += 1
     if crr%1000 == 0:
         print(crr)
+        break
     if tokenizer.vocab[unseen] not in sum_tensors.keys():
         unseen_tokens.append(unseen)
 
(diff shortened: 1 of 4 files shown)


MihaiSurdeanu (approved):
Thank you @RazvanDu !
LGTM. @kwalcock ?

main/src/main/python/embedding_generator/main.py
```diff
@@ -0,0 +1,119 @@
...
+
+while tokens_so_far < number_of_tokens:  # 10 million
+
+    index = np.random.randint(0, len(dataset))
+
+    text = dataset['train'][index]['text']
+
+    tokenized = tokenizer(text, return_tensors='pt')
+
+    tokens = tokenized['input_ids'][0]
```
kwalcock, line 58: So is [0] for the first sentence, the first word, first layer?

main/src/main/python/embedding_generator/main.py
```diff
@@ -0,0 +1,119 @@
...
+        tokens_so_far += len(tokens)
+
+        sliced_output = {key: value[:, :] for key, value in tokenized.items()}
+
+    else:
+
+        i = np.random.randint(0, len(tokens) - context - 1)
+
+        crr += 1
+        tokens_so_far += context
```
kwalcock, line 72: I don't understand what tokenizer() returns or how this subsampling of tokenized.tiems works, but I'll take your word for it.

kwalcock (approved):
Thanks for the extensive documentation.  I'm not sure whether my comments are worth worrying about.

main/src/main/python/embedding_generator/main.py
```diff
@@ -0,0 +1,119 @@
...
+crr = 0
+
+sum_tensors = {}
+crr_tensors = {}
+
+i = 0
+
+print(len(dataset))
+
+while tokens_so_far < number_of_tokens:  # 10 million
```
kwalcock, line 50: Is this condition here because the input will likely have too many tokens to do them all once?  With the small test set, it means that they will all be done multiple times, perhaps unnecessarily.  The RNG might cause some indices to be visited multiple time and some none.  Could a slice of the data be used instead of the random samples with replacement?

main/src/main/python/embedding_generator/main.py
```diff
@@ -0,0 +1,119 @@
...
+
+    index = np.random.randint(0, len(dataset))
```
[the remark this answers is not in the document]
RazvanDu: Ah, I forgot about a random seed, I can add that soon.

As per the holes, repeating indices often result in completely different contexts, for reasonable-sized data sets, thus I don't think skipping or repeating tokens is much of an issue. I can look into that if you want though.

main/src/main/python/embedding_generator/main.py
```diff
@@ -0,0 +1,119 @@
...
+if huggingface_dataset:
+    if huggingface_subpart:
+        dataset = load_dataset(huggingface_dataset, huggingface_subpart).with_format(type="torch")
+    else:
+        dataset = load_dataset(huggingface_dataset).with_format(type="torch")
+else:
+    dataset = load_dataset('json', data_files={'train': 'text_data.json'})
+
+tokens_so_far = 0
+crr = 0
```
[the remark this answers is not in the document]
RazvanDu: It's a leftover from a previous metric that I was displaying, I removed it, thanks!

main/src/main/python/embedding_generator/main.py
```diff
@@ -0,0 +1,119 @@
...
+
+i = 0
+
+print(len(dataset))
+
+while tokens_so_far < number_of_tokens:  # 10 million
+
+    index = np.random.randint(0, len(dataset))
+
+    text = dataset['train'][index]['text']
```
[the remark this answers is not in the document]
RazvanDu: Usually yes, that's the way wikipedia is structured.

main/src/main/python/embedding_generator/main.py
RazvanDu: If I remember correctly the tokenizer returns multiple values, [0] takes the actual tokens.

main/src/main/python/embedding_generator/main.py
RazvanDu: For reasonable data sets, the probability of repeating a token is exceedingly small. Considering that I'd say we need at least 10M tokens for a decent coverage. With a context size of 1024 that would mean we only look at roughly 10000 indices from our 10M total indices. The actual tokens might repeat but they will most likely be part of different contexts, thus having different embeddings.

RazvanDu:
Alright I fixed a small typo issue, remove the useless 'crr', and I added seeds. Let me know if there's anything else to implement.

kwalcock:
Thanks for adding the seed.  I did look up [numpy.random.seed](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) and you might want to read it in case it is ever important.  If, for example, any of the dependencies sets the seed itself or for some reason generates an extra random number because it is Tuesday, your reproducibility might not work.  It might be worth documenting that -1 is a special value that means not to set the seed at all.

For crr I was just wondering what the letters mean.  Google didn't help me with it.

I'll ask again about the sampling in person and in the meantime check this in.  Change anything if/when you like, of course.

RazvanDu:
> Thanks for adding the seed.  I did look up [numpy.random.seed](https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html) and you might want to read it in case it is ever important.  If, for example, any of the dependencies sets the seed itself or for some reason generates an extra random number because it is Tuesday, your reproducibility might not work.  It might be worth documenting that -1 is a special value that means not to set the seed at all.
> 
> For crr I was just wondering what the letters mean.  Google didn't help me with it.
> 
> I'll ask again about the sampling in person and in the meantime check this in.  Change anything if/when you like, of course.
> 
> 

Oh that's interesting, I'll look into it soon.

Crr is a short form of current, I use it when I write code in a rush/very fasy.

Alright, thanks!

PR #766, opened 2023-12-12
Update dependencies

kwalcock:


kwalcock:
This PR updates dependencies to the currently published scala-transformers and models.  There should be three failing tests as described shortly in an issue.  They would need to be fixed/disabled before balaur can be merged into master.

PR #769, opened 2023-12-20
Trace what is happening with simple example rules and document

kwalcock:
See #761 first


kwalcock:
@navalani, you may be interested in this.

navalani (approved):

PR #772, opened 2024-01-08
Put clulab in charge of its own holidays

kwalcock:


MihaiSurdeanu:
Nice, thank you @kwalcock !

kwalcock:
I think this is good enough for now.  It gives us access to newer holidays even though it uses an older JollyDay, and we can (and have) added our own holidays.  The HolidayNormalizer should be multiple times faster because it doesn't go through holidays of every separate state and it matches based on keys rather than descriptions, which are an extra call away.

kwalcock:
P.S. This should prevent eidos from crashing while processing habitus files and SEC software from crashing on interviews.

MihaiSurdeanu (approved):
Thanks!

PR #777, opened 2024-02-04
Date rule issue776

MihaiSurdeanu:


main/src/main/resources/org/clulab/numeric/WEEK.tsv
```diff
@@ -5,12 +5,24 @@
 # note: multi-word phrases must be tokenized in the same way as our tokenizer. If not sure, try the phrases in ./shell first!
 #
 first week       // XXXX-XX-01 -- XXXX-XX-07
-1st week       // XXXX-XX-01 -- XXXX-XX-07
+1st week         // XXXX-XX-01 -- XXXX-XX-07
```
kwalcock, line 8: I'm not sure this is what it really means, but that might have been discussed at length previously.  If Saturday is the 1st, then the first week might be 3-9 for me.  Haven't I seen it written -W1?  The first week of June is XXXX-06-W1.

kwalcock: I will work on #778 and tidy a few things below and merge it all at once if that's OK.

kwalcock:
OK.  Also, the Germans and Dutch number the weeks of the year, so 2023-W46, aside from the numbering within the month.  There's a rule about what the first week is based on which weekday New Year's falls on.  Maybe that's where I got it from.  We probably never run into it in English text.

Oops.  Google tracked this down: https://www.epochconverter.com/weeknumbers.  Maybe it is a broader thing.  If the Thursday rule/guideline holds within months, it could be used to produce exact dates when the year and month are known or a -W when not.

kwalcock (approved):

PR #781, opened 2024-02-06
Get NumericEntityRecognizerShell to work again

kwalcock:


kwalcock:
This change was added to a branch off the branch used to produce v9.0.0-RC4.  The change could be ported back to the balaur branch, which is dependent on snapshots and doesn't have all the other changes that were necessary for other versions of Scala or have updates that had been made to master.  It is probably time to call the branch here masterV9 (or mainV9) and continue using it rather than balaur.   Exercising it would be a good thing in case I have introduced bugs not yet discovered.

MihaiSurdeanu:
I agree. Thank you!

PR #782, opened 2024-02-07
Add visualize to the Inst trait

kwalcock:


kwalcock:
@navalani, these couple of additional rules are sufficient to provide simple examples of objects of the other Inst subclasses.

navalani:
Thanks Keith, I will modify the drawing to include those subclasses

kwalcock:
@navalani, hopefully the computer can soon make the visualizations itself.

navalani (approved):

navalani:
Hey Keith,

I noticed in the code that we are currently visualizing the part of the
rule that is matched, i.e (B-Food), (I-Food), etc. but do we need to
visualize which part of the sentence that matches or is the next.getPosId
sufficient?



Nick

kwalcock:
Hi,

The first step is to show the B-Food or I-Food in a "visualization" so that
the user can understand which Inst is in question.  At this point we are
still not matching against any text like in a sentence.  The step after
that is to show what the B-Food or I-Food turned out to match (or failed to
match) in the sentence like the cake, pain au chocolat (or John eats).

Keith

navalani:
I see. Thanks for the clarification. Will work on that right now.



Nick

navalani:
Do you have any suggestions on where to look for what matches the rule?
For example, for MatchToken, I traced down the classes that extend
TokenConstraint and StringMatcher but to no avail. They only hold [B-Food]
and/or [I-Food] and not where the rule matches.



Nick

kwalcock:
@navalani, you will eventually need the tok field of the Thread trait.  I believe this keeps track of what token (of a sentence) is being matched against which Inst (of the rule).  We might eventually keep a list of tokens that did or didn't match each Inst so that the user can verify that their rule is working as expected.  For the moment we're still wanting to make sure the user and computer understand the rule in the same way based on the visualization of the Insts before they even start working.  One thing that we're debugging is an incorrectly stated rule that doesn't mean what the rule writer thinks it does.  We're more or less showing the assembly language that their code (rule) produced with enough clues so that they can verify it or at least do a sanity check.

navalani:
Understood. Thanks

navalani:
I just pushed some code that added a visualization that shows what B-Food
and B-Per match in MatchToken. Before I add visualization for the other
Inst classes, I wanted to make sure, the tok field in Thread gives the
index of the token in the sentence right? For example, tok = 0 in "John
eats cake" refers to John. The visualization I pushed works with that
understanding.


Nick

kwalcock:
@navalani, it's cool to see the "B-FOOD matches cake".  That will give the user some indication of what is happening.   The new code in ThompsonVM.stepSingleThread will probably become relevant soon.  Things happening in that area are probably more complicated than OdinStarter would leave one to believe.  Try the sentence `John eats cake and Jane eats pain au chocolat.` for a preview of the complications.  That B-FOOD will end up matching `pain` and the record of `cake` will have been lost.  The `val matchTokens = new HashMap[String, Int]` will need to be a more complicated mechanism for recordkeeping.  It will additionally need to keep track of the sentence, because the one that is being worked on in stepSingleThread is not necessarily that one that is being used to make the visualization in OdinStarter (i.e., mentions.head.sentenceObj).  However, that's a worry for later.

In the very short term, can you adjust the recursion in OdinStarter so that the output does not keep repeating the name of the extractor (rule)?  Right now it is
```
There was an extractor: foods-from-lexicon - Inst: 1. SaveStart(--GLOBAL--)
1. SaveStart(--GLOBAL--)
There was an extractor: foods-from-lexicon (Next) - Inst: 2. MatchToken(B-FOOD matches pain) -> 3
2. MatchToken(B-FOOD matches pain) -> 3
There was an extractor: foods-from-lexicon (Next) (Next) - Inst: 3. Split.  Check out my LHS and RHS!
There was an extractor: foods-from-lexicon (Next) (Next) (LHS) - Inst: 4. MatchToken(I-FOOD matches chocolat) -> 3
4. MatchToken(I-FOOD matches chocolat) -> 3
There was an extractor: foods-from-lexicon (Next) (Next) (RHS) - Inst: 5. Pass
There was an extractor: foods-from-lexicon (Next) (Next) (RHS) (Next) - Inst: 6. SaveEnd(--GLOBAL--)
6. SaveEnd(--GLOBAL--)
There was an extractor: foods-from-lexicon (Next) (Next) (RHS) (Next) (Next) - Inst: 0. Done
```
but it would be much more understandable like
```
There was an extractor: foods-from-lexicon
1. SaveStart(--GLOBAL--)
2. MatchToken(B-FOOD matches pain)
3. Split.  Check out my LHS and RHS!
    (LHS) 4. MatchToken(I-FOOD matches chocolat) -> 3
    (RHS) 5. Pass
    6. SaveEnd(--GLOBAL--)
    0. Done

There was an extractor: person-from-lexicon
1. SaveStart(--GLOBAL--)
etc.
```
or something similar.  Numbering, spacing, reference to other numbers would all be very useful.

navalani:
Just pushed some code that has a more understandable visualization. I tried
"John eats cake and Jane eats pain au chocolat." and I see what you mean.
The record of cake is lost. Let me try using a HashMap with an array of
Integers as the value.




Nick

kwalcock:
@navalani, the updated formatting looks much better (IMO).  A novice rule writer can probably learn fairly quickly to follow it.  I do notice that some consecutive numbers are missing and that the Done has disappeared, so some logic is not yet quite right.  If this was less experimental, unit tests would be written to vouch for its correctness, but for now it's just eyeballs.

If you are already looking at the HashMap that is stored in the ThompsonVM, I would suggest moving it to the Debugger class instead.  It would also be good to take another look at #769.  You might notice that each Inst gets run/processed/visited numerous times.  I see debugDoc, debugLoop, debugExtractor, debugSentence, debugTokenInst.  These Doc, Loop, Extractor, Sentence, TokenInst form a context in which an Inst might match a token or not.  I am leaning towards making something like that essentially the key to your HashMap and performing queries on it.  It would be mini database.  After that `extractorEngine.extractFrom(document)` in OdinStarter, one would say to the Debugger, possibly for the sake of the visualization, but maybe more generally, "You just extracted from a document.  I'm interested in that 5th sentence.  There's that foods-from-lexicon rule that didn't seem to work right.  The B-FOOD Inst is acting up.  Which tokens did it match?"  This is similar to what happens in
```scala
  for (extractor <- extractors)
    visualize(extractor, sentence)
```
but for a more complex and realistic situation.  There's more to say about it, perhaps at a meeting.  @MihaiSurdeanu will likely have an opinion based on experience (while I'm just guessing :-)

PR #791, opened 2024-03-15
Show one way of recording virtual machine operations

kwalcock:


kwalcock:
@navalani, there are some debugger ideas here, but I wouldn't necessarily use them directly in your code, so this is a draft.  You can however read it or check it out and try it or copy parts.

One thing it does is show in the output is a trace the recursive search for matching Insts, which has helped me understand what is happening.  Another thing it does is keep track of the circumstances/context that applies when making the match.  See the DebuggerRecord for that.  Some of these additional fields would eventually need to be tracked in your PartialMatch class.  I am using dumb records here and I can imagine the PartialMatch doing a smarter job if you can figure out when to add things to it.  Probably wherever I call debugMatch because something matched or didn't, you would want to update the partial matches if it does.  I may not have all the right spots yet for lack of testing.

There is code in OdinStarter with examples of the kinds of questions that can be answered with the transcript.

```
Rule foods-from-lexicon matched sentence "John and Jim eat [pain au] chocolat but not pizza .".
Rule foods-from-lexicon matched sentence "John and Jim eat [pain au chocolat] but not pizza .".
Rule person-from-lexicon matched sentence "[John] and Jim eat pain au chocolat but not pizza .".
Rule person-from-lexicon matched sentence "John and [Jim] eat pain au chocolat but not pizza .".
Rule people-eat-food matched sentence "John and Jim [eat] pain au chocolat but not pizza .".

Rule foods-from-lexicon partially matched sentence "John and Jim eat pain au chocolat but not >pizza< .".
Rule more-person-from-lexicon partially matched sentence "><John and Jim eat pain au chocolat but not pizza .".
```

navalani:
I see. Will keep this in mind when I implement the Partial Matches class.

PR #796, opened 2024-04-25
Print matchLookAhead and Behind in OdinStarter

kwalcock:


kwalcock:
@navalani, these lines might be missing.

kwalcock:
@navalani.  They might need to print out something more complete.

navalani:
I see. I'll make those changes. By more complete, do you mean the token offsets at which they match?

kwalcock:
@navalani, I was wrong.  I see the "hello"  come up in the next Inst.

kwalcock:
@navalani, there is additional code to store more than one match per Inst.

kwalcock:
@navalani, since Insts are identified only by the posId, they need to be compared in a different way.  It was an optimization.

kwalcock:
@navalani, I notice that visualization code has returned to OdinStarter.  At one time it was moved to the Debugger.  The duplication may be causing you or the computer some confusion.

navalani:
Right. Let me work on transferring the code over.

navalani:
visualize() and visualizeExtractor() are found both in the OdinStarter and in the Debugger, I am assuming we want them only in the Debugger and calls made to them from the Odin starter

navalani:
Got it. Thanks

kwalcock:
@navalani, FWIW I don't see that the IdentityHashMap from the ThompsonVM made it into your code.  I would guess that's why au is being overwritten by chocolat.  It looks like all four As are matching.  Git is difficult to figure out and I think some things are getting lost, some things that might have been helpful.

navalani:
You're right, it'll be useful to run diff with the two versions of the Thompson VM. Let me work on that

navalani:
@kwalcock is the IdentityHashMap in the ThompsonVM in the debugContext branch? I don't see much difference in the files

kwalcock:
@navalani, it should be in this branch, kwalcock/odin-debugger, from which this PR originates.

PR #801, opened 2024-07-11
MAJOR reorganization of processors

MihaiSurdeanu:
We now support only in-house tools implemented in BalaurProcessors.


MihaiSurdeanu:
@kwalcock: to the best of my knowledge, this is ready to become the new master and released.
Thank you!

kwalcock:
@MihaiSurdeanu, I notice warnings about the apps being in the org.clulab.processors package which is where org.clulab.processors.Processor is.  This would have a package being split between jars, which can lead to problems.  It'd like to move the apps to org.clulab.processors.apps.

kwalcock:
@MihaiSurdeanu, there are more apps within the library project, many of which depend on a BalaurProcessor which would not work without the model.  Should they be moved to the apps project?

Is there enough left in the library project for it to be useful without the model?  Is there particular functionality that is targeted?

org.clulab.sequences.LexiconNERShell, OK
org.clulab.utils.ProcessCoNLL03, has Balaur
org.clulab.numeric.NumericEntityRecognizerShell, has Balaur
org.clulab.odinstarter.OdinStarter, has Balaur
org.clulab.processors.clu.VeilApp, has Balaur
org.clulab.utils.MEDApp, OK
org.clulab.processors.apps.ExtractSentencesApp, has Balaur

kwalcock:
The hexatags with tl-label, tr-label, nl-label, nr-label, or eos are examples of string typing.  I'm supposed to give people a hard time about that.  Things like
```scala
HexaTags.NONTERM_LEFT_CHILD + "-" + label
termTags(0) = termTags(0).filter(_._1.startsWith(HexaTags.TERMINAL_LEFT_CHILD))
 ```
might be better avoided.  I started converting, but there aren't many tests and not much documentation, so I don't think it's a good idea right now.

**Comment by MihaiSurdeanu** [comment]
Thanks @kwalcock !
Yes, the apps you mentioned should be moved.

Let's leave the strings as they are for now. I can perhaps do it, if you tell me what best practices are.

**Comment by kwalcock** [comment]
Maybe strings can be replaced in 10.1.  At one point the code looked like this below, but there are other ways to do it. 
```scala
trait HexaTag {
  val label: String

  def isLeft: Boolean
  def isRight: Boolean = !isLeft
  def isTerminal: Boolean
  def isNonTerminal: Boolean = !isTerminal
}

case class TreeHexaTag(override val label: String, override val isTerminal: Boolean, override val isLeft: Boolean) extends HexaTag {

  override def toString: String = {
    val first = if (isTerminal) "t" else "n"
    val second = if (isLeft) "l" else "r"

    s"$first$second-$label"
  }
}

case class EosHexaTag() extends HexaTag... // Is this really necessary?
```
I'm not sure about the heap in MinHeap.  It seems like that should be an ArrayBuffer.

There will be a PR soon for the apps.

kwalcock:
There are several other files that include `def main`.  Three make use of the processor and have to be moved.  I was going to leave the other ones where they are.  One included comments insisting to do that.

```
CoNLLToSentencePerLine, OK
LearningExample, OK
RankerEvaluator, OK
EvalTimeNorm, has Balaur
DepsToTags, OK
DepsToTagsApp, OK
LexiconSentimentAnalyzer, has Balaur
NormalizeParens, OK
```

MihaiSurdeanu:
Look to move the ones with Balaur. Thank you!

PR #802, opened 2024-07-12
Prepare for release

kwalcock:


kwalcock:
Still tracking down apps...

MihaiSurdeanu (approved):
Thanks @kwalcock !

kwalcock:
There's some kind of issue I need to figure out.  BTW Jenkins had been down after the river crash from Wednesday or maybe earlier.  It is back up since Friday.

MihaiSurdeanu:
I wonder if we could switch to GitHub Actions at this moment? I **think** that they offer more free compute than Travis?

kwalcock:
The problem is that the jenkins image that is being used does not have some libraries that components from rust need.  System libraries are missing.  The same problem might happen for users of the processors.  They go to run it and find out that they need to install libssl and libcrypto, which has not been documented.

kwalcock:
This is being closed in order to clean up Jenkins

PR #804, opened 2024-07-15
Create ci.yml

kwalcock:


kwalcock:
Everything seems to imply that the change must be made to master before any branches or other PRs can be checked.

PR #807, opened 2024-07-16
Prepare for release

kwalcock:


kwalcock:
This finally passed the tests.  The problem wasn't with the code but with the configuration of the test computer.

kwalcock:
It looks like this is OK to merge into hexatags.  If it doesn't cause problems there, hexatags could go to master.  I wonder if this would be a good time to switch from master to main before it gets published.

MihaiSurdeanu (approved):
Looks great, thank you @kwalcock !

PR #809, opened 2024-07-22
Show alternative form of webapp

kwalcock:


kwalcock:
For typical clulab webapps that implement a simple REST api like this, without user management, lots of state, distributed deployment, etc., the play framework is excessive and has lots of disadvantages like the need for sbt plugins, a completely different project structure, very different deployment procedure.  This is an example of using a web server library, of which there are plenty, to create a simple app that implements the api.  There's a WebApp class here to specify the interface and a WebProcessor to implement it.  Assembly results in a processors-webapp jar file that can be run with `java -jar processors-webapp.jar`. in an otherwise empty directory.  This is how the habitus clustering server will work.  This doesn't need to be merged.  It's just FYI and the PR will be deleted.

https://github.com/clulab/processors/blob/kwalcock/http4s/webapp2/src/main/scala/org/clulab/processors/webapp/WebApp.scala

https://github.com/clulab/processors/blob/kwalcock/http4s/webapp2/src/main/scala/org/clulab/processors/webapp/WebProcessor.scala

Issue #812, opened 2024-09-29
These sentences cause processors to crash


kwalcock:
The problem was noted with version 8.5.4 and they crash the currently deployed webapp.  These have to do with numbers.  A rule finds a mention, but the processing of the mention doesn't work.  There may be exception handling in place that should result in the mention being thrown away, but it doesn't seem to be working.

```
pdf 75 Some of the major sectoral plans include the national energy policy, transport policy, forest and wildlife policy, national environmental sanitation strategy etc. 76 https://www.thegasconsortium.com/documents/GMP-Final-Jun16.pdf 77 http://www.energycom.gov.gh/files/Renewable-Energy-Masterplan-February-2019.pdf 78 http://energycom.gov.gh/files/SE4ALL-GHANA%20ACTION%20PLAN.pdf 79 http://www.energycom.gov.gh/files/Ghana%20Integrated%20Power%20System%20Master%20Plan%20_Volume%202.pdf Page | 80 In 2017, the Ministry of Energy started a comprehensive review of the National Energy Policy, which is still ongoing.
```

```
16 http://www.energycom.gov.gh/files/Renewable-Energy-Masterplan-February-2019.pdf 17 http://energycom.gov.gh/files/SE4ALL-GHANA%20ACTION%20PLAN.pdf Page | 14 Forestry National Forest and Wildlife Policy Stumpage Fees Surcharge on timber as part of the timber harvesting regulation regime.
```

```
They are: * One-district one factory, * Integrated Aluminium Industry * Planting for food and jobs, * One village one dam * Aquaculture for food and jobs * One district one warehouse Table 3 shows the linkages between nationally determined contributions, national climate change policy, medium-term development frameworks, government flagships projects and SDGs 18 https://www.mwh.gov.gh/wp-content/uploads/2018/05/SECTOR-MEDIUM-TERM-DEVELOPMENT-PLAN-2018-2021.pdf 19 http://nadmo.gov.gh/images/NADMO_documents/2015_documents/GHANA%20PLAN%20OF%20ACTION%20ON%20DRRCCA%202011-2015.pdf 20 Essebey G.O, Nutsukpo D, Karbo N, and Zougmore R. 2015.
```

```
[...]

PR #814, opened 2024-10-06
Fail better on percentages

kwalcock:


kwalcock:
See #812.  This may not be the correct fix, but it does seem to take care of the exception in the right way.  This fix will be propagated to master eventually.

MihaiSurdeanu:
Thank you @kwalcock ! This helps a lot. It look like I will have no time in the next couple of weeks for coding...

kwalcock:
See also #815

kwalcock:
I'm probably going to go ahead and merge this into this old branch first.  I think this was necessary for a publishLocal that I was using for something.  Then I will fix it up again for a PR to master, which is right now showing merge conflicts.

Issue #816, opened 2024-10-11
More sentences crash processors


kwalcock:
This seems to have problems unrelated to the other ones:

```
<img data-cke-saved-src=”https://ichef-1.bbci.co.uk/news/624/cpsprodpb/A28E/production/_96741614_f2f1e164-0deb-4b81-a586-66ac3804fe66.jpg” src=”https://ichef-1.bbci.co.uk/news/624/cpsprodpb/A28E/production/_96741614_f2f1e164-0deb-4b81-a586-66ac3804fe66.jpg” datasrc=”https://ichef-1.bbci.co.uk/news/320/cpsprodpb/A28E/production/_96741614_f2f1e164-0deb-4b81-a586-66ac3804fe66.jpg” class=”responsive-image__img js-image-replace” alt=”MSNBC Public Relations tweets: ” it’s=”” a=”” sad=”” day=”” for=”” america=”” when=”” the=”” president=”” spends=”” his=”” time=”” bullying,=”” lying=”” and=”” spewing=”” petty=”” personal=”” attacks=”” instead=”” of=”” doing=”” job.
```

kwalcock:
False alarm.  It was failing the numbers check and was thrown out.

kwalcock:
However, it crashes the present webapp, so there must be something more that is problematic about it.

kwalcock:
The problem is that it fails in HexaDecoder.decode(...).  For each loop it throws the exception that is ignored (https://github.com/clulab/processors/blob/2c3b3f75992bc95cfe3d5640cb0bc1c5174bffcd/library/src/main/scala/org/clulab/processors/hexatagging/HexaDecoder.scala#L59) and so the dependencies don't get filled in.  The webapp and other applications should be prepared for this, but perhaps it also shouldn't be happening.

PR #817, opened 2024-10-18
Account in the webapp for missing dependencies

kwalcock:
An alternative solution would be to return Some(empty directed graph) rather than None


kwalcock:
See #816

kwalcock:
These are changes to the webapp to account for the possibility of dependencies not being there.  Other clients of the library would have to account for this as well.  FYI Mihai.

MihaiSurdeanu:
Nice. Thank you @kwalcock !

MihaiSurdeanu:
We should release 10.1.0 once these PRs are merged.

PR #819, opened 2024-12-02
Adjust merge strategy

kwalcock:


kwalcock:
You might need to replace your local mainclass branch with the version from the server.

MihaiSurdeanu:
Thank you @kwalcock !

PR #828, opened 2025-01-22
Work around some bugs in numeric processing

kwalcock:


kwalcock:
These are the same changes that were added to a different branch in #814.  The code needed to fail at a different time.

kwalcock:
@MihaiSurdeanu, that's a very good point.  I'll check that it works for both.

kwalcock:
There is then a related issue with
```
"forty-one" | "forty-two" | "forty-three" | "forty-four" | "forty-five" | "forty-six" | "forty-seven" | "forty-eight" | "forty-nine" |
```
which don't offer the alternate spelling.

MihaiSurdeanu:
> There is then a related issue with
> 
> ```
> "forty-one" | "forty-two" | "forty-three" | "forty-four" | "forty-five" | "forty-six" | "forty-seven" | "forty-eight" | "forty-nine" |
> ```
> 
> which don't offer the alternate spelling.

Yeah, I think it's worth expanding this regex as well.

kwalcock:
@MihaiSurdeanu, this was updated.

MihaiSurdeanu:
Thank you @kwalcock ! Merge?

kwalcock:
@MihaiSurdeanu, if it looks OK now.

PR #832, opened 2025-03-05
Add debugger subproject

kwalcock:


MihaiSurdeanu:
Nice!

kwalcock:
@MihaiSurdeanu, documentation is at https://github.com/clulab/processors/wiki/Debugging and https://github.com/clulab/processors/blob/kwalcock/collapse/docs/debugging.md, but I still need to fill in the links (or get rid of them).  I will try a PR against the main branch as well.

PR #842, opened 2025-05-27
Modernize with immutable Document and Sentence

kwalcock:


kwalcock:
One disadvantage is that access to the Scala Seqs from Java is more complicated than it is to Array.  This shows especially in JavaProcessorsExample.scala (https://github.com/clulab/processors/blob/kwalcock/processors2b/apps/src/main/java/org/clulab/processors/apps/ProcessorsJavaExample.java)

kwalcock:
@MihaiSurdeanu, it is partly to enforce it (immutability), but almost just as much to document it.  Scala 2.13 and 3 compel one to do this by not allowing without significant complaint and making copies of the arrays in cases like

```scala
val seq: Seq[Int] = Array(1, 2)
```
which is what all the implicit conversions in e.g. https://github.com/clulab/processors/tree/kwalcock/processors2b/library/src/main/scala-3/org/clulab/scala are meant to avoid.  This update uses the newer ArraySeq from Scala 2.13 and Scala 3 to handle the issue and backports to Scala 2 using scala-collection-compat.  A more thorough update would make the other implicit conversions unnecessary, but that's too big a step just now.

library/src/main/scala/org/clulab/processors/clu/BalaurProcessor.scala
```diff
@@ -34,230 +35,248 @@ class BalaurProcessor protected (
...
-        sent.words,
-        sent.tags,
-        sent.lemmas,
-        entities = None,
-        norms = None,
-        chunks = None,
-        tree = None,
-        deps = EMPTY_GRAPH,
-        relations = None
+        words, // TODO: Why isn't this raw?
```
kwalcock, line 213: Here is a worrisome TODO.

library/src/main/scala/org/clulab/processors/clu/BalaurProcessor.scala
```diff
@@ -34,230 +35,248 @@ class BalaurProcessor protected (
...
         )
-      } catch {
-        case e: EncoderMaxTokensRuntimeException => 
-          // this sentence exceeds the maximum number of tokens for the encoder
-          // TODO: at some point do something smart here
-          println(s"ERROR: this sentence exceeds the maximum number of tokens for the encoder and will not be annotated: ${sent.words.mkString(" ")}")
 
+        partlyAnnotatedSentence
+      }
+      // TODO: Improve error handling.
```
kwalcock, line 169: I'd like to do something better here.

library/src/main/scala/org/clulab/processors/clu/BalaurProcessor.scala
MihaiSurdeanu: I vaguely remember this... I think some lexicon NER dictionaries operate over lemmas rather than words, which required such weird work arounds...

MihaiSurdeanu (approved):
Thank you @kwalcock !

kwalcock:
The ColumnsToDocument code included at least two places where it is implied that a function manipulates a sentence or document:
```scala
setLabels(s, labels.toArray)

val d = new Document(sentences.toArray)
annotate(d)
```
That doesn't work anymore, so the function signatures were changed.  ColumnsToDocument is in the apps subproject even though it isn't an application.  It is not used anywhere in the project that I can find.  Is it then library code that some client is expected to make use of?  Perhaps it should be moved to the library subproject.

MihaiSurdeanu:
> The ColumnsToDocument code included at least two places where it is implied that a function manipulates a sentence or document:
> 
> ```scala
> setLabels(s, labels.toArray)
> 
> val d = new Document(sentences.toArray)
> annotate(d)
> ```
> 
> That doesn't work anymore, so the function signatures were changed. ColumnsToDocument is in the apps subproject even though it isn't an application. It is not used anywhere in the project that I can find. Is it then library code that some client is expected to make use of? Perhaps it should be moved to the library subproject.

I used it I think when I created the training data for the parser. That can be easily adjusted.

PR #844, opened 2025-07-08
Show how to keep track of Odin rule source

kwalcock:
This is built on top of other draft PRs, but could be isolated and added without them.  Some changes are made to Odin and regression tests are advised, because local unit tests may not be adequate to detect subtle behavior changes.  The only thing that should affected is some debugger output.  The key addition is of the SourcingParsers class and that's where the secret sauce lies.  The last commit includes some refactoring which may not be desired and could be removed.


MihaiSurdeanu (approved):
Thank you @kwalcock !!

